PR #1, opened 2022-06-27
Resampling + To_Relocations functions

cath34:
Examples are available at the end of the Relocations and Trajectories notebook.


ecoscope/base/base.py
```diff
@@ -516,6 +518,165 @@ def turn_angle(traj):
 
         return angles.rename("turn_angle").reindex(self.index)
 
+    def upsample(self, upsample_time):
```
kulits, line 521: Consider renaming `upsample_time` to `freq` and changing type from `int` to `pd.Timedelta` to match pandas.

ecoscope/base/base.py
```diff
@@ -516,6 +518,165 @@ def turn_angle(traj):
...
+
+            traj_ind = traj_ind.sort_values("segment_start")
+            times = pd.date_range(
+                start=traj_ind.iloc[0]["segment_start"],
+                end=traj_ind.iloc[-1]["segment_end"],
+                freq=pd.Timedelta(seconds=upsample_time),
+            )
+            index = traj_ind.segment_start.searchsorted(
+                pd.date_range(
+                    traj_ind.segment_start.iat[0],
```
kulits, line 552: Appears to be a duplicate of `times`.

ecoscope/base/base.py
```diff
@@ -516,6 +518,165 @@ def turn_angle(traj):
...
+        -------
+        ecoscope.base.Relocations
+        """
+
+        relocs_gdfs = []
+
+        def compute(traj):
+            traj.crs = self.crs
+            subject_name = traj.name
+            points = np.concatenate(
```
kulits, line 593: Replace with speedier `pygeos` function: `pygeos.get_point(trajectory.geometry.values.data, 0)`.

ecoscope/base/base.py
```diff
@@ -516,6 +518,165 @@ def turn_angle(traj):
...
+            traj_ind.to_crs(utm_crs, inplace=True)
+            logging.info(f"Resampling relocations for subject {subject_name}")
+
+            traj_ind = traj_ind.sort_values("segment_start")
+            times = pd.date_range(
+                start=traj_ind.iloc[0]["segment_start"],
+                end=traj_ind.iloc[-1]["segment_end"],
+                freq=pd.Timedelta(seconds=upsample_time),
+            )
+            index = traj_ind.segment_start.searchsorted(
```
kulits, line 550: What about resampled times that don't fall within a `Trajectory` segment (multiple collars)? I think you might need a second searchsorted along with filtering.

ecoscope/base/base.py
cath34: It's not

ecoscope/base/base.py
cath34: The function works with resampled times that don't fall within a trajectory segment (see tests for an example). I also think it can handle big gaps in the data, such as re-collaring ones. I haven't test it properly tho but I can make a synthetic examples to see if it works

ecoscope/base/base.py
cath34: Fixed

ecoscope/base/base.py
cath34: I think it's easier for a user that doesn't have much programming knowledge to use seconds (int) than a Timedelta object.

ecoscope/base/base.py
cath34: Okay, fixed now

PR #5, opened 2022-10-06
Added transform parameter to the to_geotiff function

cath34:


notebooks/03. Home Range & Movescape/EcoGraph.ipynb
```diff
@@ -69,7 +70,7 @@
     "    from google.colab import drive\n",
     "\n",
     "    drive.mount(\"/content/drive/\", force_remount=True)\n",
-    "    output_dir = os.path.join(\"/content/drive/MyDrive/\", output_dir)\n",
+    "    output_dir = os.path.join(output_dir, \"/content/drive/MyDrive/\")\n",
```
walljcg, line 73: this is backwards and was correct before- the output_dir should be appended to the /content/drive/MyDrive base folder

walljcg (changes requested):
This is great - thanks. I just had one comment about the output_dir folder path for Google Drive. And also, I don't think we should be committing test outputs to the repo: tests/test_output/salif_degree_mean_std.feather

notebooks/03. Home Range & Movescape/EcoGraph.ipynb
kulits: Please do not hard code Google Drive paths in the notebooks. They should be able to be run locally.

cath34: > This is great - thanks. I just had one comment about the output_dir folder path for Google Drive. And also, I don't think we should be committing test outputs to the repo: tests/test_output/salif_degree_mean_std.feather

The test_output folder contains files that allow you to verify that the output of some Ecoscope functions are those you would expect. I added a test that checks if the function output is the appropriate transformed GeoTIFF, so I added the expected output (in feather format) to the test_output folder in order to check with geopandas.testing if the function output was okay

walljcg: Ok, yes that makes sense

walljcg (approved):

PR #9, opened 2022-10-23
Modify text and position of Historic Mean

ericgitonga:


walljcg (approved):
Looks good, thanks Eric!

PR #11, opened 2022-10-25
fix cmap error value

ericgitonga:
"turbo" is not a valid cmap value. It brought up an error. Switched to "jet"


kulits:
https://github.com/matplotlib/matplotlib/pull/15275

PR #14, opened 2022-10-26
`is_monotonic` is deprecated

zakhusayn:
**FutureWarning**: `is_monotonic` is deprecated and will be removed in a future version. Use `is_monotonic_increasing` instead.


walljcg (approved):
Thanks Zak. Sorry this was left for so long!

PR #18, opened 2022-11-15
Bump to version 1.1.0

ericgitonga:


walljcg (commented):
Hi Eric, I think let's use the convention that we will increment versions according to: 

1.x.y  where y increases when there are minor changes to existing functionality or bug fixes. Then x increments when we introduce new functionality or major changes to existing functionality. And then the major version, e.g., v1 -> v2 increments when we have major incompatible backward changes or significant changes to project structure.

PR #19, opened 2022-12-03
Groupby col

ericgitonga:
Make the patrols relocs use the patrol_id as the groupby-col instead of the subject_id


ecoscope/version.py
```diff
@@ -1 +1 @@
-__version__ = "1.0.1"
+__version__ = "1.1.0"
```
walljcg, line 1: I think let's bump to 1.0.2 instead as I see this as a bug fix rather than introduction of new functionality

walljcg (changes requested):

ecoscope/version.py
ericgitonga: This bump was for the previous additions to Ecoscope... I had made the PR to bump to 1.1.0 before I did this change of patrol_id. So for patrol_id, it ought to be bumped to 1.1.2...

ericgitonga: I have bumped the version to 1.1.1

walljcg: Thanks

walljcg (approved):

PR #20, opened 2022-12-06
Ecograph fix

cath34:
Fixed a bug + Prevention of warnings than can occur when computing an EcoGraph on huge trajectories/huge dataset

### DIFF
--- a/ecoscope/analysis/ecograph.py
+++ b/ecoscope/analysis/ecograph.py
@@ -1,396 +1,693 @@
-from math import ceil, floor
+import warnings
 
+import astroplan
+import astropy
 import geopandas as gpd
-import igraph
-import networkx as nx
 import numpy as np
 import pandas as pd
-import rasterio
-import sklearn.base
-from affine import Affine
-from shapely.geometry import shape
-from skimage.draw import line
+import pygeos
+from pyproj import Geod
 
-import ecoscope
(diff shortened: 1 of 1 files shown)


ecoscope/analysis/ecograph.py
```diff
@@ -197,7 +197,10 @@ def _get_ecograph(self, trajectory_gdf, individual_name, radius, cutoff, tortuos
 
         for node in G.nodes():
             for key in attributes.keys():
-                G.nodes[node][key] = np.mean(G.nodes[node][key])
+                if len(G.nodes[node][key]) != 0:
```
walljcg, line 200: i think this would be simpler using https://numpy.org/doc/stable/reference/generated/numpy.nanmean.html

walljcg (changes requested):

ecoscope/analysis/ecograph.py
cath34: You are right : Using np.nanmean prevents the code from checking for nan values in other functions. I changed it. However, we still to keep the if len(G.nodes[node][key]) != 0, because it's possible for a given node to have no parameter values at all. In that case, using np.mean or np.nanmean will throw the warning « computing mean of empty slice » before returning np.nan. In order to prevent that warning, we need to check the length of the node attribute values.

walljcg (approved):

PR #22, opened 2022-12-21
ECO-130

cath34:


ecoscope/io/earthranger.py
```diff
@@ -708,6 +579,28 @@ def get_events(
         df.sort_values("time", inplace=True)
         return df
 
+    def get_patrols(self, filter=None, status=None, **addl_kwargs):
+        """
+        Parameters
+        ----------
+        filter:
```
walljcg, line 586: Can we expand the single filter parameter to multiple params in the signature? I find the use of a filter here confusing compared with our other function definitions

ecoscope/io/earthranger.py
```diff
@@ -708,6 +579,28 @@ def get_events(
...
+            example: {"date_range":{"lower":"2020-09-16T00:00:00.000Z"}}
```
walljcg, line 587: Think we should provide since and until parameters here rather than a single date range parameter which is harder for the user to construct.

ecoscope/io/earthranger.py
```diff
@@ -842,11 +724,9 @@ def post_subjectsource(
         lower_bound_assignend_range
```
walljcg, line 724: spelling mistake here (assigned)

ecoscope/io/earthranger.py
```diff
@@ -922,3 +798,94 @@ def patch_event(
...
             results = [self._patch(f"activity/event/{event_id}", payload=events)]
         return pd.DataFrame(results)
+
+    def post_subject(
+        self,
+        subject_name: str,
+        subject_type: str,
+        subject_subtype: str,
+        is_active: bool = True,
+        tracks_available: bool = True,
```
walljcg, line 808: tracks_available might be an internal parameter. I would suggest we don't let the user try and set this.

ecoscope/io/earthranger.py
```diff
@@ -922,3 +798,94 @@ def patch_event(
...
+
+        response = self._post("subjects", payload=payload)
+        return pd.DataFrame([response])
+
+    def post_source(
+        self,
+        source_type: str,
+        manufacturer_id: str,
+        model_name: str,
+        provider: str,
```
walljcg, line 844: does this require the provider ID value? Can it be null?

ecoscope/io/earthranger.py
```diff
@@ -922,3 +798,94 @@ def patch_event(
...
+            "provider": provider,
+        }
+
+        if kwargs:
+            payload.update(kwargs)
+
+        response = self._post("sources", payload=payload)
+        return pd.DataFrame([response])
+
+    def post_patrol(self, priority: int, **kwargs) -> pd.DataFrame:
```
walljcg, line 875: I see that patrols consist of patrol segments. Suppose we need a function for posting those as well? Are there any other parameters that would define a patrol, e.g., title etc that we can expose to the user?

ecoscope/io/earthranger.py
```diff
@@ -187,38 +114,25 @@ def get_subjects(
...
-            Indicate whether to render each subject's last location.
-        tracks
-            Indicate whether to render each subject's recent tracks.
-        id
-            A comma-delimited list of Subject IDs.
+        subject_group: Indicate a subject group for which Subjects should be listed.
+        name : Find subjects with the given name [UUID]
+        updated_since: Return Subject that have been updated since the given timestamp.
+        render_last_location: Indicate whether to render each subject's last location.
+        tracks: Indicate whether to render each subject's recent tracks.
```
walljcg, line 126: I think we should set this to false and hide from user as an param. I think this  param applies mainly to the web app.

ecoscope/io/earthranger.py
```diff
@@ -187,38 +114,25 @@ def get_subjects(
...
-        updated_since
-            Return Subject that have been updated since the given timestamp.
-        render_last_location
-            Indicate whether to render each subject's last location.
-        tracks
-            Indicate whether to render each subject's recent tracks.
-        id
-            A comma-delimited list of Subject IDs.
+        subject_group: Indicate a subject group for which Subjects should be listed.
+        name : Find subjects with the given name [UUID]
```
walljcg, line 123: this is confusing - do we want a string name or an uuid here?

ecoscope/io/earthranger.py
```diff
@@ -187,38 +114,25 @@ def get_subjects(
...
-        include_inactive
-            Include inactive subjects in list.
-        tracks_since
-            Include tracks since this timestamp
-        tracks_until
-            Include tracks up through this timestamp
-        bbox
-            Include subjects having track data within this bounding box defined by a 4-tuple of coordinates marking
+        include_inactive: Include inactive subjects in list.
+        tracks_since: Include tracks since this timestamp
```
walljcg, line 118: think we should hide this param.

ecoscope/io/earthranger.py
```diff
@@ -187,38 +114,25 @@ def get_subjects(
...
+        tracks_until: Include tracks up through this timestamp
```
walljcg, line 119: think we should hide this param

ecoscope/io/earthranger.py
```diff
@@ -1,172 +1,99 @@
...
-            geometry=gpd.points_from_xy(df["location"].str[longitude], df["location"].str[latitude]),
-            crs=4326,
-        )
+    @backoff.on_exception(backoff.expo, requests.exceptions.RequestException, max_tries=10, giveup=fatal_status_code)
+    def _get(self, path, stream=False, **kwargs):
+        headers = {"User-Agent": self.user_agent}
+        headers.update(self.auth_headers())
 
-    @staticmethod
-    def _normalize_column(df, col):
```
walljcg, line 136: do we not need this function anymore?

walljcg (changes requested):
Thanks for your work on this Catherine. Have made a few comments.

ecoscope/io/earthranger.py
cath34: Oops, that's an error. I put it back

ecoscope/io/earthranger.py
cath34: You're right. I removed the tracks_available parameter.

ecoscope/io/earthranger.py
cath34: Fixe

ecoscope/io/earthranger.py
cath34: I added a post_patrol_segment function. Yes, there are also other parameters that can define a patrol (objective, title, files, notes). All of them can be specified through kwargs. I guess that title could be added as an optional parameter for the post_patrol function since it's a quite common patrol parameter?

ecoscope/io/earthranger.py
cath34: No : only the name. Provider is a required parameter to create a source from the API, but I modified the function so that "default" is the provider name if the user does not specify it.

ecoscope/io/earthranger.py
cath34: The given string name. I made it more precise

PR #26, opened 2023-04-21
Update EarthRangerIO from DAS Client to ER Client

ericgitonga:


tests/test_earthranger_io.py
```diff
@@ -76,14 +77,14 @@ def test_get_subjectsource_no_observations(er_io):
     assert relocations.empty
 
 
-def test_get_subjectsource_observations_with_pagesize_one(er_io):
+def test_get_subjectsource_observations_with_cursor(er_io):
```
walljcg, line 80: I think this goes away as we're not using cursors?

walljcg (changes requested):
Thanks Eric for the massive branch of work! Looks good other than dropping the test using a cursor.

tests/test_earthranger_io.py
ericgitonga: I have removed the cursor test.

So far, all tests for earthranger_io.py run successfully save for the test_get_observation_for_patrol. And I think that's still due to my difficulty in posting observations to patrol segments.

The other issues I had faced were:

Fixing the bug that resulted in us not being able to get subjects by group_name. The way we were accessing the required id was wrong, thus the error we kept getting of "id" not an attribute of type "list"

Fix test_users(): We needed to get the output of er_io.get_users() into a dataframe to properly test whether it was empty or not.

I had to remove the test_get_subject_no_observations() test. When you try pull subject observations with a non-existent subject id, you get back a 404 error. Even on the ER interactive API online. This behaviour is different from pulling source observations. With that, if you pass in a non-existent source id, you get back a 200 response, but with no data. This is an ER API issue from the looks of it. So unless that is fixed, that particular test is useless.

For the test_get_events() test, initially we passed in page_size=100 to it, causing the test to complain that there were multiple values of page_size passed to it. This is because we are doing the same when we call get_objects_multithreaded(). I removed that, but then now had to change the test from asserting to check for len(events) <= 100 to asserting to ensure events is not empty.

There are 2 other pending failures, but those are for test_ud.py and test_utils.py. There are also 6 errors reported for test_ecodataframe.py

The only other thing that's pending is updating the EarthRanger IO notebook, but I'd want to update that after fixing the patrol segment events issue. Unless you think otherwise.

Unless there's something else I may have overlooked, er_client_ is now good to go.

ericgitonga: Oh, there was one other test_earthranger_io failure which was caused by wrong naming of event_type in MEP-Dev. Editing that event type fixed the error. This was for the patch_event test.

PR #39, opened 2023-06-20
Replacing pygeos with shapely

ericgitonga:


walljcg (approved):
Looks good, thanks Eric

PR #42, opened 2023-08-09
Add get_patrols_from_id() function

ericgitonga:


walljcg (changes requested):
The naming of this function is confusing to me. Maybe call it get_patrols_from_patrol_type_id  so that we're explicit?

ericgitonga:
> The naming of this function is confusing to me. Maybe call it get_patrols_from_patrol_type_id so that we're explicit?

I get your confusion, and on looking at the code again, I realise that get_patrols_from_patrol_type_id is not the right name either. We are pulling patrol segments from particular patrol IDs, and it is from those segments that we then extract the events. A fitting name would therefore be get_patrol_segments_from_patrol_id.

walljcg (approved):

PR #43, opened 2023-08-17
add-north-arrow-flexibility

gcastellini:
-Added feature to select an image instead of default SVG.
-Added angle option.
-Modified ControlElement class.


walljcg (commented):
It seems like the control element class is being overridden here to support the North Arrow functionality only. But we will be working with other elements that may require different behaviour. Would it be better to create a SubClass of ControlElement specifically (e.g.) NorthArrowElement that we then extend with properties such as angle, image etc.?

walljcg (approved):
Nice. Thank-you.

PR #44, opened 2023-09-02
return event types as df and use event_id as index in get_events

ericgitonga:
Changes made to earthranger.py to have the dataframe returned by get_events() use the event_id as the index, and to have get_event_types() return a dataframe.


walljcg (commented):
What does the include_schema parameter do? Do you know what the schema is?

ericgitonga:
> What does the include_schema parameter do? Do you know what the schema is?

As per the API, it is supposed to return the schema specified in an event types definition. When you run the command through the interactive API, you get the schema returned. However, when you run through the ER  code, that doesn't happen. It looks like it is just an informational parameter that doesn't serve any real need as far as I can tell. It is, therefore, safe to remove it.

walljcg (approved):

PR #45, opened 2023-09-04
Fix chromedriver dependency

gcastellini:
-Modified to_png method to include Pyppeteer library
-Removed chromedriver dependency.


walljcg (approved):
Looks good!

PR #53, opened 2023-10-26
chore: fix docs compilation.

enlivn:
# Purpose
chore: fix docs compilation.

# Considerations
- Doc compilation has been broken since #16 because the environment was not solvable. Updated versions of the libraries to a state where docs compile again.
- I've updated the branch protection rules for `master` to prevent any further pulls into `master` from being mergeable unless the checks are green. That way, this shouldn't happen again.
- tests are passing so I'm assuming this set of library versions is still functionally correct. @ericgitonga, any chance you could double-check that this set of lib versions works okay on Colab? I don't have access to it.
- this change is needed so I can create another PR that updates docs for Ecoscope GUI
- 8 tests have been failing since May 2022. I've disabled them so we can get the builds green again. They'll be fixed in a follow-up PR against a JIRA ticket (I don't have the number since I'm still waiting on JIRA acess).

# Successful build
image

https://readthedocs.org/projects/ecoscope/builds/22349258/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,8 +1,9 @@
 version: 2
 build:
-  os: ubuntu-20.04
+  os: ubuntu-22.04
   tools:
-    python: mambaforge-4.10
+    python: mambaforge-22.9
+
 sphinx:
   configuration: doc/source/conf.py
 python:
--- a/environment.yml
+++ b/environment.yml
@@ -2,46 +2,47 @@ name: ecoscope
 channels:
   - conda-forge
 dependencies:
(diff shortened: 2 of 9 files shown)


enlivn:
Current dependencies on/for this PR:
* master
  * **PR #53** <a href="https://app.graphite.dev/github/pr/wildlife-dynamics/ecoscope/53" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wildlife-dynamics/ecoscope/53?utm_source=stack-comment).

walljcg (approved):
I've tested this in Colab and is working.

PR #64, opened 2023-12-06
remove remaining collar voltage code from core library

ericgitonga:


ecoscope/plotting/__init__.py
```diff
@@ -12,7 +11,6 @@
 __all__ = [
     "EcoPlotData",
     "add_seasons",
-    "collar_event_timeline",
```
walljcg, line 15: is this moved to a module now?

ecoscope/plotting/plot.py
```diff
@@ -158,42 +150,6 @@ def add_seasons(fig, season_df):
     return fig
 
 
-def collar_event_timeline(relocations, collar_events):
```
walljcg, line 161: where has this function moved to?

walljcg (changes requested):

ecoscope/plotting/__init__.py
ericgitonga: I did not include it, but can do so. Question, when is it used? I haven't seen an example where it has been called.

ecoscope/plotting/plot.py
ericgitonga: I have moved the function to the Collared Subject Report notebook, in the collar event cell. I have tested it to make sure it runs just fine.

PR #67, opened 2024-01-02
Eco 260 ee geemap update

ericgitonga:
Added code from the geemap project to initlaize EarthEngine. Also updated setup.py and environment.yml to include basic minim libraries need for functional ecoscope install.


ecoscope/contrib/geemap.py
```diff
@@ -0,0 +1,96 @@
+import ee
```
walljcg, line 1: Will need to copy over the current MIT license supplied by geemap. Take a look at the foliummap contrib module we have for the example.

ecoscope/io/eetools.py
```diff
@@ -95,12 +97,7 @@ def label_gdf_with_img(gdf=None, img=None, region_reducer=None, scale=500.0):
 
 
 def _match_gdf_to_img_coll_ids(
```
walljcg, line 99: I prefer having the function params on separate lines which makes tracking changes to function signatures easier in github, and also prevents exceeding pep8 requirements for longer functions.

setup.py
```diff
@@ -14,32 +14,27 @@
...
     "matplotlib",
     "networkx",
     "numba",
-    "numexpr",
-    "numpy",
-    "pandas",
     "plotly",
+    "pre-commit",
     "pyarrow",
-    "pyproj",
```
walljcg, line 35: sure we don't need pyproj?

walljcg (changes requested):

setup.py
ericgitonga: I have tested setting up the environment several times over with this setup.py file, and that has not been an issue. That said, I can leave it in, since having it in place doesn't affect the environment.

ecoscope/contrib/geemap.py
ericgitonga: Done. I also moved out the code from __init.py__ to geemap and did away with importing the version and function from there so we have all geemap code in just one file as well, so we don't have to replicate the license info.

ecoscope/io/eetools.py
ericgitonga: Looking through the Black documentation, it says you can use this to have it ignore formatting:

# fmt: off

block of code

# fmt: on

Thing is, you have to do this for each section you want it to ignore. At least as of what I have found from my brief search. I am looking to see if there is a way we can universally have it ignore function definitions so we don't have to individually ignore them in the code.

walljcg: Ok interesting. Well, it's not so important that I think we should turn off that formatting rule. Let's stick with it for now. Thanks for looking into it.

walljcg:
@ericgitonga ok, I think you can merge and close. Thanks

walljcg (approved):

PR #69, opened 2024-01-13
fix get_events() time bug

ericgitonga:
I have skipped the failing tests/test_earthranger.py for get_events(). It is choking on one of the time formats, so I need to explore why that is.

This fix is to enable production to function as I look at the failing test.


walljcg (approved):
Thanks Eric, but please add as Jira ticket to work on in next sprint (i.e., why this was failing?)

PR #91, opened 2024-04-05
fix bug in _normalize_column() that resulted in error

ericgitonga:
.iteritems() has been deprecated, and now we need to use .items() instead.


walljcg (approved):
Thanks for fixing these

Issue #96, opened 2024-04-06
Method to download files


walljcg:
Our current implementation is here: https://github.com/wildlife-dynamics/ecoscope/blob/d0650725ed91d2cb1d6b3decc24be3486302dd00/ecoscope/io/utils.py#L28

Need to make this more generalized to support download of files: 

- [ ] Dropbox
- [ ] Google Drive
- [ ] Zipped files (and unzip them)

- explore using fsspec as used by GeoPandas?

Yun-Wu:
Done

PR #110, opened 2024-04-09
Overhaul download_files and support unzipping

Yun-Wu:
#96 

1. Add unittest for downloading data from Github/Google Drive/Dropbox
2. Support unzipping 
3. Fix unittest discovery failure
4. Update environment.yml to include new dependencies


tests/test_io_utils.py
```diff
@@ -0,0 +1,57 @@
...
+        output_dir,
+        overwrite_existing=True,
+    )
+
+    data = pd.read_csv(os.path.join(output_dir, "movbank_data.csv"))
+    assert len(data) > 0
+
+
+def test_download_file_dropbox_json():
+    URL = "https://www.dropbox.com/scl/fi/qaw3krcsnot69x94mdfxy/config.json?rlkey=zdmipl2la7rplgl218vc13end&dl=1"
```
walljcg, line 48: Think we should include code to automatically change dl=0 to dl=1 to help out users from having to modify a dropbox url themselves.

tests/test_io_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    data = pd.read_csv(os.path.join(output_dir, "download_data.csv"))
+    assert len(data) > 0
+
+
+def test_download_file_gdrive():
+    output_dir = "tests/test_output"
+    ecoscope.io.download_file(
+        "https://drive.google.com/uc?export=download&id=1-AQ9_oacUCcAaiZ6SWU77hZWp1oArQw6",
```
walljcg, line 26: This url string differs from what a user gets when they click on a GDrive file and select 'Copy Link'. Can we transform from that into what's needed here for downloading? E.g., https://docs.google.com/document/d/1EeBSLv7LOr6maTTVY8T08oD0zbsMb9iUyrAjetIWYUo/edit?usp=sharing

tests/test_io_utils.py
```diff
@@ -0,0 +1,57 @@
...
+        os.path.join(output_dir, "download_data.csv"),
+        overwrite_existing=True,
+    )
+
+    data = pd.read_csv(os.path.join(output_dir, "download_data.csv"))
+    assert len(data) > 0
+
+
+def test_download_file_gdrive_zip():
```
walljcg, line 35: Is it possible to combine with the above function so that it works for both non-zip and zip files?

walljcg (changes requested):

tests/test_io_utils.py
Yun-Wu: It works for both non-zip and zip files. Here are just two different unittests

tests/test_io_utils.py
Yun-Wu: done

ericgitonga (commented):
All tests pass fine.

PR #145, opened 2024-05-08
Workflows cleanup: use secrets directly and dedupe test runs

cisaacstern:
Closes #143

This PR cleans up our CI testing / workflows developer experience in two ways:
1. Removes use of a namespaced Deployment "environment" for holding our GitHub Secrets in favor of using the global GitHub Secrets namespace directly. This aligns us with the most common usage pattern for testing CI wherein the test checks are nested under a collapsible status icon associated with a _commit_, and there is not (as we currently have) a cluttering of the (non-collapsible) PR history with Deployments notifications (as shown in the screenshot in #143).
2. Also I've changed the triggering rules to de-duplicate test runs for PRs: currently we are getting a run for the commit to the PR _and_ a duplicate run for the push to the branch from which the PR originates. By changing the rule to only trigger test runs on pushes to the default branch name, this duplication will be avoided.

Note: To get CI to work with this new secrets configuration, we will need to recreate the Testing Deployment Secrets as Repository Secrets, on https://github.com/wildlife-dynamics/ecoscope/settings/secrets/actions, here:
Screenshot 2024-05-08 at 11 58 36 AM
For probably good security reasons, we can't move existing secrets, so they will need to be re-added by someone who has access to the original secrets.

PR #141, opened 2024-05-06
[SPIKE] Distributed functions

cisaacstern:
WIP based on this morning's pairing sessions with @walljcg.

I will clean this up and add tests before marking as ready for review.

PR #148, opened 2024-05-08
[SPIKE] Distributed tasks

cisaacstern:
Supersedes #141 and includes the CI changes from #145. Re-opening a new PR because IIUC once a PR is opened, the "expected checks" can't (or at least sometimes won't) change. So I think the only way to get the CI changes from #145 reflected here is to re-open this. Copying remaining TODOs from a comment on #141:

- [x] decorator-based serialization/deserialization handling
- [x] ensure decorator works as expected when wrapped by pydantic validate_call
- [ ] importable (in backend) with minimal dependencies
- [ ] end-to-end example (time density) including data loading and results model (with serialization between steps)


cisaacstern:
Superseded by https://github.com/wildlife-dynamics/ecoscope-workflows/pull/2

PR #170, opened 2024-05-24
Implement feature density function

Yun-Wu:
- Add point tests
- Add line test
- remove raster code


ecoscope/io/raster.py
```diff
@@ -204,3 +206,34 @@ def raster_to_gdf(raster_path):
             ],
             crs=src.crs,
         )
+
+
+def grid_to_raster(grid=None, val_column="", out_dir="", raster_name="grid.tif", xlen=5000, ylen=5000, grid_crs=4326):
+    """
+    Save a GeoDataFrame grid to a raster.
+    Does not handle different xlen and ylen
```
atmorling, line 214: Does this mean that the xlen and ylen params must be equal? If so why not a single param?

Yun-Wu: I tested different `xlen` and `ylen` and they are handled differently. It may be an outdated comment from the copy-over code.

atmorling (approved):
Approved. 

Note to self, will have to update the test here when merging #178

PR #193, opened 2024-06-29
Numba jit-compilation for intersect1d loop in calculate etd

cisaacstern:
Based on @atmorling's observations in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/45#issuecomment-2197203917, made an attempt to optimize this loop. This code doesn't work yet, but just throwing it up as draft so others can see, and perhaps offer some insights!

Not our highest priority right now, but I was too curious to not try. 😄

### DIFF
--- a/ecoscope/analysis/UD/etd_range.py
+++ b/ecoscope/analysis/UD/etd_range.py
@@ -76,6 +76,42 @@ class Weibull3Parameter(WeibullPDF):
     c: float = 1.0
 
 
+@nb.jit(
+    nb.float64[:](
+        nb.int32,
+        nb.int32,
+        nb.types.Tuple((nb.float64[:], nb.float64[:])),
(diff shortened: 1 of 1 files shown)


walljcg:
I wonder if JAX would help at all (when we go down the rabbit hole)  :)

cisaacstern:
Via offline review from @alxmrs:
> "If you have a slow Set intersection operation, consider using a bloom filter. JIT on an O(N) algorithm won’t help much; you need a better data structure. A skip list might be a good fit, but I think you’ll really enjoy reading up on and implementing a bloom filter."

Alex, hope it's okay to post that review here, wanted to make sure others on this thread benefitted from your insights!

atmorling:
Closing this in order to clean up the PR queue, I've raised #373 to track any ongoing efforts here but this is firmly a story for another day.

PR #199, opened 2024-07-05
Get patrols and their observations in the same method

Yun-Wu:


Yun-Wu:
https://github.com/wildlife-dynamics/ecoscope-workflows/issues/42

atmorling (approved):
Thanks Yun

PR #207, opened 2024-07-16
Fix unittest failure

Yun-Wu:


atmorling (approved):
lgtm, thanks for fixing this!

PR #233, opened 2024-08-01
Add dependabot.yaml to auto-update github actions versions

cisaacstern:
This will instruct Dependabot to automatically open PRs like #232 so we don't need to do this manually.

See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


atmorling (approved):
LGTM!

PR #245, opened 2024-08-09
Ecoscope 197

davidovichmarcos:
Adding tests to execute notebooks with papermill on the CI


.github/workflows/_tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Run tests
+on:
+  push:
+    branches: [ "*" ]
+env:
+  ER_USERNAME: ${{ secrets.ER_USERNAME }}
+  ER_PASSWORD: ${{ secrets.ER_PASSWORD }}
+  GCP_DATAFLOW_SERVICE_KEY: ${{ secrets.GCP_DATAFLOW_SERVICE_KEY }}
```
cisaacstern, line 8:
```suggestion
  GEE_SERVICE_ACCOUNT_KEY: ${{ secrets.GEE_SERVICE_ACCOUNT_KEY }}
```

Let's rename this env variable and the corresponding secret to something more descriptive. If this is only used for Google Earth Engine (which I think it is), then maybe `GEE_SERVICE_ACCOUNT_KEY`?

(For others reading this, our reference for this workflow was a prior workflow I'd worked on for a previous job, in which we were deploying to GCP Dataflow from GitHub. But we're not doing that here, so the naming should reflect what we're actually doing.)

Also, @davidovichmarcos, how did you generate the secret used here? Is it in fact a service account or is this key tied to your personal user account? If the latter, let's make an actual service account for this.

.github/workflows/_tests.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  run_unit_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v1'
+        with:
+          credentials_json: '${{ secrets.GCP_DATAFLOW_SERVICE_KEY }}'
```
cisaacstern, line 17:
```suggestion
          credentials_json: '${{ secrets.GEE_SERVICE_ACCOUNT_KEY }}'
```

.github/workflows/_tests.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install dependencies
+        run: pip install -r requirements-notebooks-test.txt
+      - name: Use the kernel
+        run: python -m ipykernel install --user --name=venv
+      - name: Run unit tests
+        run: pytest nb-tests/test_notebooks.py -s
+    env:
+      ER_USERNAME: ${{ secrets.ER_USERNAME }}
+      ER_PASSWORD: ${{ secrets.ER_PASSWORD }}
+      GCP_DATAFLOW_SERVICE_KEY: ${{ secrets.GCP_DATAFLOW_SERVICE_KEY }}
```
cisaacstern, line 36:
```suggestion
      GCP_DATAFLOW_SERVICE_KEY: ${{ secrets.GEE_SERVICE_ACCOUNT_KEY }}
```

nb-tests/test_notebooks.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+
+
+NB_DIR = pathlib.Path(__file__).parent.parent / "doc" / "source" / "notebooks"
+
+KNOWN_ERRORS_REGEXES = {  # This is basically a GitHub ticket queue
+    "EarthRanger_IO.ipynb": "Series found",
+    "GEE_IO.ipynb": "frontend does not support input requests", # This error only happens in the ci pipeline
+    "EcoMap.ipynb": "frontend does not support input requests", # This error only happens in the ci pipeline
+    "Remote Sensing Time Series Anomaly.ipynb": "frontend does not support input requests", # This error only happens in the ci pipeline
```
cisaacstern, line 15: Do you think we can resolve the errors marked as

```python
"frontend does not support input requests"  # This error only happens in the ci pipeline
```
before merging?

IMHO, it makes sense to not fix _all_ "KNOWN ERRORS" before merging, since many of these are not problems specific to the CI pipeline. But if the error itself if specific to the CI pipeline, I think it makes sense to fix that here. One reason being is that the notebooks failing in this way may have other "actual" errors in them which are not being caught because this error is being thrown first.

If I had to guess, these errors are being thrown because we are not providing the correct environment variables to the Earth Engine initialization? And therefore we are hitting this block

https://github.com/wildlife-dynamics/ecoscope/blob/3caa5e6b754e119af263f1ee0ec7dd706db83ae9/ecoscope/contrib/geemap.py#L127-L129

which presumably is asking for user input to move forward?

Optimistically, I'd think the fix is "just" a matter of translating the GEE credentials we are using in the workflow into the correct format to be picked up here:

https://github.com/wildlife-dynamics/ecoscope/blob/3caa5e6b754e119af263f1ee0ec7dd706db83ae9/doc/source/notebooks/01.%20IO/GEE_IO.ipynb#L91-L92

?

I would assume `EE_PRIVATE_KEY_DATA` is _probably_ the `json.load`'ed representation of `GEE_SERVICE_ACCOUNT_KEY`. And the `EE_ACCOUNT` is the username associated with that account?

There is a possibility once these variables are parsed correctly, that we will find that the secret configured here is not in fact registered with GEE. (That is, I'd guess once "frontend does not support input requests" is fixed, we will see something like "This account does not exist on GEE".) In which case we will need to resolve that as well.

.github/workflows/_tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Run tests
+on:
+  push:
+    branches: [ "*" ]
```
cisaacstern, line 4:
```suggestion
  push:
    branches: [ "master" ]
  pull_request:
    branches: [ "*" ]
```

I think we should use the same trigger as in `main.yaml` for uniformity.

.github/workflows/_tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Run tests
```
cisaacstern, line 1:
```suggestion
name: Run notebook tests
```

I think this file is just notebook tests, correct?

cisaacstern (changes requested):
@davidovichmarcos thanks so much for getting this so close, overall looks great.

I did take the somewhat "extreme" stance of setting this review as "Request changes" because there are a few things that I think we should clean up here before merge.

In addition to the comments below, I think we should rename `_tests.yml` to something like `notebook-tests.yml` which is (a) more descriptive; and (b) doesn't start with underscore (because it is running directly, not being inherited).

Let me know if I can clarify anything. Again, amazing work, this will be a huge leap forward here! 👏

.github/workflows/_tests.yml
atmorling: There's an actual service account for this configured in the repo settings already but it's stored as `EE_ACCOUNT` & `EE_PRIVATE_KEY_DATA` which _may_ be compatible here with some changes to the yaml.

atmorling: @davidovichmarcos I can share those credentials with for local testing

atmorling (commented):
Thanks for sorting this out @davidovichmarcos :pray: 
The only thing I'd add in addition to the credentials configuration is a preference to  to make this more aligned with the existing workflows which have an `environment.yaml` / `micromamba` setup, rather than `requirements.txt`

As I was writing this I realised we probably want to do away with both of these approaches and do something simpler per the CI in `ecoscope-workflows`. Have created #247 to track that and leaving this here just to complete the loop. No action required here from you @davidovichmarcos

.github/workflows/_tests.yml
davidovichmarcos: I'm using this service account keys (which i created only for that purpose) but i can try setting a new one, and yes i'm using a real json file with valid credentials.

Screenshot 2024-08-16 at 9 49 48 AM

requirements-notebooks-test.txt
```diff
@@ -0,0 +1,8 @@
+pip-tools
+pytest
+papermill
+folium
+branca
```
cisaacstern, line 5:
```suggestion
```

cisaacstern (approved):
Amazing!!!

Now that we have mapping deps coming from ecoscope directly, I don't think we need folium and branca (see below).

cisaacstern:
🚀 👏

PR #264, opened 2024-09-08
fix ER test, get_subjects now retuns is_static

atmorling:


atmorling:
[or does it?](https://github.com/wildlife-dynamics/ecoscope/actions/runs/10762638716/job/29843221048#step:4:401)

PR #269, opened 2024-09-11
Allow classifier to label ranges

### LINKED ISSUE #267: `apply_classification` should allow cusomisation of label values
In a case such as classifying continuous values into discrete bins, it would be helpful to allow a user to specify how the label values are formatted, without needing to know the exact ranges of bins as created by `apply_classify`

atmorling:
Updates `apply_classification` to support generating label ranges if the classification values are numerical.
Also supports label prefix/suffixes

closes #267


ecoscope/analysis/classifier.py
```diff
@@ -75,7 +85,19 @@ def apply_classification(
     classifier = classifier_class(dataframe[input_column_name].to_numpy(), **kwargs)
     if labels is None:
         labels = classifier.bins
+
+        # Generate range labels if our bins are numeric
+        if np.issubdtype(dataframe[input_column_name].dtype, np.number) and label_ranges:
+            # We could do this using mapclassify.get_legend_classes, but this generates a cleaner labely
+            ranges = [f"0 - {labels[0]:.{label_decimals}f}"]
```
walljcg, line 92: One concern here is if the min value of the classified value is not zero? E.g., if we classify NDVI values into bins but their value might run from -0.85 to 0.95 for example.

ecoscope/analysis/classifier.py
```diff
@@ -75,7 +85,19 @@ def apply_classification(
     classifier = classifier_class(dataframe[input_column_name].to_numpy(), **kwargs)
     if labels is None:
         labels = classifier.bins
+
+        # Generate range labels if our bins are numeric
+        if np.issubdtype(dataframe[input_column_name].dtype, np.number) and label_ranges:
+            # We could do this using mapclassify.get_legend_classes, but this generates a cleaner labely
```
walljcg, line 91: "but this generates a cleaner label"

ecoscope/analysis/classifier.py
```diff
@@ -25,7 +26,16 @@
...
+    labels=None,
+    scheme="natural_breaks",
+    label_prefix="",
+    label_suffix="",
+    label_ranges=False,
+    label_decimals=1,
+    **kwargs,
 ):
     """
     Classifies the data in a GeoDataFrame column using specified classification scheme.
```
walljcg, line 41: "Classifies the data in a DataFrame column using specified classification scheme."

walljcg (changes requested):

ecoscope/analysis/classifier.py
atmorling: I can remove this so it will always be `dataframe[input_column_name].min()`

atmorling: Done - have replaced 0 with `dataframe[input_column_name].min()`

atmorling:
@walljcg  would you be able to give the changes here a look when you have a chance?

atmorling:
closes #267

davidovichmarcos (approved):

Yun-Wu (approved):

PR #272, opened 2024-09-15
Add timestamp to patrol events

Yun-Wu:


ecoscope/io/earthranger.py
```diff
@@ -683,6 +714,10 @@ def get_patrol_events(self, since=None, until=None, patrol_type=None, status=Non
         events_df = clean_time_cols(events_df)
 
         events_df["geometry"] = events_df["geojson"].apply(lambda x: shape(x.get("geometry")))
+        events_df["time"] = events_df["geojson"].apply(
+            lambda x: datetime.datetime.strptime(x.get("properties").get("datetime"), "%Y-%m-%dT%H:%M:%S%z")
+        )
+
```
atmorling, line 720: I feel like we eventually (not necessarily with this PR) want this logic to be a utility function like `clean_time_cols` but for parsing the geojson we get from EarthRanger. 

I'm not sure if it's something that _can_ happen within EarthRanger, but I think it's worth surfacing an error message that indicates bad/missing data if these lambdas were to fail. Rather than the `AttributeError` or `TypeError` that would materialism currently. 

I don't think this needs to be addressed with this PR, just writing this here for posterity.

atmorling (approved):
LGTM, Just the one comment that I don't think needs to be addressed in this PR - I've raised https://github.com/wildlife-dynamics/ecoscope-workflows/issues/240 to track

ecoscope/io/earthranger.py
Yun-Wu: Yea, I agree. Thanks Alex!

PR #275, opened 2024-09-30
Improve arguments of get_patrols

Yun-Wu:
- **format since and until timestamp**
- **add invalid format**
- **Support patrol type values**
- **comment**


atmorling (approved):
LGTM!

PR #302, opened 2024-10-20
Add NSD Calculation

### LINKED ISSUE #280: Add NSD to trajectory calculation
I think we used to have this but seems to have been dropped. Would like to add the Net-Square Displacement calculation when we generate a trajectory. The basis for the logic is here: 

```python
def calculate_nsd(traj):
    def calculate_nsd(group):
        first_point = group.geometry.iloc[0]
        group['nsd'] = group.geometry.apply(lambda point: 
             point.distance(first_point) ** 2)
        return group

    nsd_df = traj.groupby('groupby_col').apply(calculate_nsd)
return nsd_df
```

Note that this one of those calculations where it only makes sense in the context of a single subject's dataset and not across multiple subjects.

Yun-Wu:
Closes #280


Yun-Wu:
The generated plot looks like this.

Yun-Wu:
The generated plot looks like this. 
截屏2024-10-20 下午5 00 36

walljcg (approved):
Looks great, thanks!

PR #316, opened 2024-11-15
coerce all parsed timestampts to UTC

### LINKED ISSUE #315: `clean_time_cols` fails to mixed TZ timestamp strings into `.dt`-accessible columns
This function

https://github.com/wildlife-dynamics/ecoscope/blob/8aa05b58f706f1d84c83ceeb6a044b854c987ae6/ecoscope/io/earthranger_utils.py#L39-L45

is used to prepare dataframe time columns for being accessible via the `.dt` accessor, e.g. here

https://github.com/wildlife-dynamics/ecoscope/blob/8aa05b58f706f1d84c83ceeb6a044b854c987ae6/ecoscope/io/earthranger.py#L303-L305

If the input data has mixed timezone offsets, however, this function does correctly parse the column values into `pd.Timestamp`s, however the column itself remains of `type: object`, which makes the `.dt` accessor raise

```
*** AttributeError: Can only use .dt accessor with datetimelike values
```

This can be reproduced by passing a column of mixed timezone timestamp strings to this function, and then attempting to access `.dt` on that column, e.g.

```python
import pandas as pd
from dateutil import parser

times = [
    "2023-01-30 11:26:13.805829-08:00",
    "2023-09-27T06:16:46.158966-07:00"
]
col = pd.Series(times, dtype="object")
my = pd.DataFrame({"time": col})
my["time"] =  my["time"].apply(lambda x: pd.to_datetime(parser.parse(x)) if not pd.isna(x) else None)
my["time"].dt
```
```
*** AttributeError: Can only use .dt accessor with datetimelike values
```
Note however that if the the timezone offsets in the input data match, this error is not raised, e.g. 

```diff
times = [
-   "2023-01-30 11:26:13.805829-08:00",
+   "2023-01-30 11:26:13.805829-07:00",
    "2023-09-27T06:16:46.158966-07:00"
]
```

atmorling:
closes #315

### DIFF
--- a/ecoscope/io/earthranger_utils.py
+++ b/ecoscope/io/earthranger_utils.py
@@ -41,7 +41,7 @@ def clean_time_cols(df):
     for col in time_cols:
         if col in df.columns:
             # convert x is not None to pd.isna(x) is False
-            df[col] = df[col].apply(lambda x: pd.to_datetime(parser.parse(x)) if not pd.isna(x) else None)
+            df[col] = df[col].apply(lambda x: pd.to_datetime(parser.parse(x), utc=True) if not pd.isna(x) else None)
     return df
 
 
--- a/tests/test_earthranger_utils.py
+++ b/tests/test_earthranger_utils.py
@@ -26,9 +26,13 @@ def df_with_times():
 
 
 def test_clean_time_cols(df_with_times):
-    cleaned = clean_time_cols(df_with_times)
+    with pytest.raises(AttributeError):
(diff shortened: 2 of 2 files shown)


cisaacstern (commented):
Thanks, Alex!

My only comment is that for completeness in addressing #315, can we actually test that the resulting column does not have the `.dt` accessor before cleaning, and does have it afterwards?

Perhaps some of the other asserts are proxies for that, but I think that doing that explicitly would be helpful.

PR #342, opened 2024-11-29
Fix crs in get_displacement

### LINKED ISSUE #343: Bug: trajectory.get_displacement()
In line: https://github.com/wildlife-dynamics/ecoscope/blob/3f00eeb355fe3069cefeee1d65e732fbbaf657d0/ecoscope/base/base.py#L368

we should not be using UTM coordinates as these don't span very far. Instead, let's use the geod.inverse_transformation() (spherical distance) as we do here: https://github.com/wildlife-dynamics/ecoscope/blob/3f00eeb355fe3069cefeee1d65e732fbbaf657d0/ecoscope/base/base.py#L633

Yun-Wu:
Closes #343

### DIFF
--- a/ecoscope/base/base.py
+++ b/ecoscope/base/base.py
@@ -367,14 +367,11 @@ class Trajectory(EcoDataFrame):
         """
         Get displacement in meters between first and final fixes.
         """
-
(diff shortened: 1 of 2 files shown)


atmorling (approved):
LGTM !

Issue #352, opened 2024-12-04
get_events isn't cleaning time_cols correctly


atmorling:
It's cleaning the raw EarthRanger df rather than the constructed gdf we expect to return

PR #357, opened 2024-12-11
Possible fix for #356

### LINKED ISSUE #356: In EarthRangerIO.get_patrol_events, geojson datetime format is not always `"%Y-%m-%dT%H:%M:%S%z"`
In querying some data from easterisland as part of https://github.com/wildlife-dynamics/ecoscope-platform-workflow-issues/issues/10, I found that the assumption of time format `"%Y-%m-%dT%H:%M:%S%z"`

https://github.com/wildlife-dynamics/ecoscope/blob/f3af43bb733a6d023ffbe17a00e001bc700dee30/ecoscope/io/earthranger.py#L755

is unsafe. As, for example, I am getting some datetimes from easterisland like

```
'datetime': '2024-12-09T21:00:11.160513+00:00'
```
which throw

```
ValueError: time data '2024-12-09T21:00:11.160513+00:00' does not match format '%Y-%m-%dT%H:%M:%S%z'
```

cisaacstern:
closes #356 

This works, but there may be a better way to do this, thus "possible" in the title.

And specifically, perhaps @atmorling has some insight into any relationship between #356 and the fix for https://github.com/wildlife-dynamics/ecoscope/issues/352 (which may point to a better solution).

### DIFF
--- a/ecoscope/io/earthranger.py
+++ b/ecoscope/io/earthranger.py
@@ -751,9 +751,22 @@ class EarthRangerIO(ERClient):
 
         events_df = clean_time_cols(events_df)
         events_df["geometry"] = events_df["geojson"].apply(lambda x: shape(x.get("geometry")))
-        events_df["time"] = events_df["geojson"].apply(
-            lambda x: datetime.datetime.strptime(x.get("properties").get("datetime"), "%Y-%m-%dT%H:%M:%S%z")
-        )
+
+        def _strptime(x: dict, fmt: str) -> datetime.datetime:
+            return datetime.datetime.strptime(x.get("properties").get("datetime"), fmt)
+
+        def _dt_from_geojson(x: dict) -> datetime.datetime:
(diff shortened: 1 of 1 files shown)


atmorling:
I think something like this makes sense:
```python
events_df["time"] = events_df["geojson"].apply(lambda x: ptime(x.get("properties").get("datetime"))
events_df = clean_time_cols(events_df)
events_df["geometry"] = events_df["geojson"].apply(lambda x: shape(x.get("geometry")))
``` 
Basically tease out the time data from the geojson _before_ we do `clean_time_cols` 
Then we shouldn't need any extra formatting in this function.

cisaacstern:
Oh cool, yeah if that works that sounds better! I figured you'd have a good insight on this.

cisaacstern:
If you'd like me to make a stab at that I can try, though tbh if you're keen, I think you're in a better place in re: familiarity with this section of code. So I'd be happy to just close this PR and have it superseded by your proposed fix.

atmorling:
Sure, will do!

PR #368, opened 2024-12-18
Add historic time series plot

### LINKED ISSUE #369: Comparative Historic Time-Series Plot
Implement this work in ecoscope core library
https://github.com/wildlife-dynamics/ecoscope-workflows/issues/397?issue=wildlife-dynamics%7Cecoscope-workflows%7C535

Yun-Wu:
Closes #369


ecoscope/plotting/plot.py
```diff
@@ -386,3 +386,98 @@ def pie_chart(
...
+            x=df.img_date,
+            y=df[current_value_column],
+            fill=None,
+            mode="lines",
+            line_color="navy",
+            name=current_value_title,
+        )
+    )
+
+    return fig
```
atmorling, line 483: Would it make sense if we had a set of style kwargs for the upper/lower bands, the historic mean and the current value? 

(I'm including `mode` because I think allowing a choice between "lines", "markers" could be useful)

Ie;
```python
def draw_historic_timeseries(
    df: pd.DataFrame,
    .
    .
    upper_lower_band_style: dict = {mode:"lines", line_color:"green"},
    historic_mean_style: dict = {mode:"lines", "line": {"color":"green", "dash":"dot")},
    current_value_style: dict = {mode:"lines", line_color:"navy"},
    layout_kwargs: dict = None,
):
    .
    .
        # add the upper bound
        fig.add_trace(
            go.Scatter(
                x=df.img_date,
                y=df[historic_max_column],
                fill=None,
                name="",
                showlegend=False,
                **upper_lower_band_style,
            )
        )
        
        # lower band
        fig.add_trace(
            go.Scatter(
                x=df.img_date,
                y=df[historic_min_column],
                fill="tonexty",
                name=historic_band_title,
                **upper_lower_band_style,
            )
        )

    if historic_mean_column:
        # add the historic mean
        fig.add_trace(
            go.Scatter(
                x=df.img_date,
                y=df[historic_mean_column],
                fill=None,
                name=historic_mean_title,
                **historic_mean_style,
            )
        )

    # add current NDVI values
    fig.add_trace(
        go.Scatter(
            x=df.img_date,
            y=df[current_value_column],
            fill=None,
            name=current_value_title,
            **current_value_style,
        )
    )

    return fig
```
Sorry I tried to do this as a suggestion but the diff was illegible :sweat:

atmorling (commented):
Just some styling comments - I'd like this to generalise beyond NDVI and I think allowing for some styling control over the individual scatter plots makes sense.

ecoscope/plotting/plot.py
Yun-Wu: Thanks for the great advise! Added them inline.

PR #375, opened 2024-12-30
Fix filter_bad_geojson

### LINKED ISSUE #376: Fix gdf error after filtered by bad json
Somehow after the bad json filter, the geodataframe becomes a regular dataframe. In `earthranger.py` line 637
```gdf.set_geometry("geometry", inplace=True)```
would fail because of this errror
```Can't do inplace setting when converting from DataFrame to GeoDataFrame```

Yun-Wu:
Closes #376


atmorling (approved):
Nice catch - Thanks! :pray:

PR #402, opened 2025-02-03
pin gee <= 1.4.5

### LINKED ISSUE #401: Temporarily pin earthengine-api
Updating the service account is blocked at the moment due to GCP permissions dependency, so will temporarily pin to a lower version of earthengine-api in order to move on

atmorling:
Closes #401


Yun-Wu (approved):
LGTM! Thanks!

PR #415, opened 2025-02-13
Fix get events

### LINKED ISSUE #414: get_events can fail if the first row has no location
Sub-task for fixing bug as described on #411

atmorling:
Closes #414


atmorling:
It was updated to reflect the scenario that caused the bug in the first place.
Previously it had data with a null `geojson` field and value in `location` which it turns out isn't true to life.

The code change as it is doesn't actually care about that difference any more, but leaving it as it's a more accurate/defensive representation of what we expect.

On another note maybe this sample should just be in geojson so changes can be diff'd here

PR #416, opened 2025-02-13
Add a general bar plot chart

### LINKED ISSUE #418: Add a general bar plot chart
Our current stacked bar charts only support time series.

Yun-Wu:
Closes #418

### DIFF
--- a/ecoscope/plotting/plot.py
+++ b/ecoscope/plotting/plot.py
@@ -345,17 +345,17 @@ def stacked_bar_chart(data: EcoPlotData, agg_function: str, stack_column: str, l
 @dataclass
 class BarConfig:
     """
-    A class to represent spec for a bar chart.
+    A class to represent configs for an individual bar in a bar chart.
     Attributes:
     ----------
     column : str
-        The name of the column to be used for the bar chart.
+        The name of the column to be used for the bar.
     agg_func : str
-        The aggregation function to be applied to the data.
+        The aggregation function to be applied to the column data.
     label : str
-        The label for the bar chart.
+        The label for the bar.
     style : dict, optional
-        A dictionary containing style options for the bar chart (default is empty).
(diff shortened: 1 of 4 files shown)


ecoscope/plotting/plot.py
```diff
@@ -340,6 +340,52 @@ def stacked_bar_chart(data: EcoPlotData, agg_function: str, stack_column: str, l
     return fig
 
 
+def bar_chart(
+    data: pd.DataFrame,
+    columns: list[str],
+    agg_funcs: list[str],
+    labels: list[str],
```
atmorling, line 347: These should always have the same length no?
Would it make sense to make this triplet a dataclass and then have a single arg?

Yun-Wu: Added BarChartSpec

ecoscope/plotting/plot.py
```diff
@@ -340,6 +342,71 @@ def stacked_bar_chart(data: EcoPlotData, agg_function: str, stack_column: str, l
...
+    Attributes:
+    ----------
+    column : str
+        The name of the column to be used for the bar chart.
+    agg_func : str
+        The aggregation function to be applied to the data.
+    label : str
+        The label for the bar chart.
+    style : dict, optional
+        A dictionary containing style options for the bar chart (default is empty).
```
atmorling, line 358:
```suggestion
    column : str
        The name of the column to be used for the bar.
    agg_func : str
        The aggregation function to be applied to the column data.
    label : str
        The label for the bar.
    style : dict, optional
        A dictionary containing style options for the individual bar (default is empty).
```

atmorling: Or something to this effect - just to make it abundantly clear that these are operating on a single bar

atmorling (approved):
LGTM outside the comment changes

PR #425, opened 2025-02-17
Filter columns from map tooltips

### LINKED ISSUE #424: Filter dataframe columns from map picking via tooltip_columns
Allow filtering of dataframe columns from picking tooltip on the created map layer

atmorling:
Closes #424


cisaacstern (approved):
LGTM!

PR #429, opened 2025-02-19
Reset df index on map layer creation

### LINKED ISSUE #428: Reset dataframe index when creating map layers
Reset a gdf's index as we `_clean_gdf` in order to avoid index information showing on the tooltip.

atmorling:
Closes #428


cisaacstern (approved):
LGTM

PR #431, opened 2025-02-20
allow patrol_type to be None

### LINKED ISSUE #430: `get_patrol_observations` can fail if the data contains patrol segments where `patrol_type` is `None`
There's an unsafe check for 'patrol_type' here:
https://github.com/wildlife-dynamics/ecoscope/blob/313cd7c2e3c61263ed377a0f79fcd2f2ef97600e/ecoscope/io/earthranger.py#L890

atmorling:
Closes #430

### DIFF
--- a/ecoscope/io/earthranger.py
+++ b/ecoscope/io/earthranger.py
@@ -887,7 +887,11 @@ class EarthRangerIO(ERClient):
                 patrol_start_time = (patrol_segment.get("time_range") or {}).get("start_time")
                 patrol_end_time = (patrol_segment.get("time_range") or {}).get("end_time")
 
-                patrol_type = df_pt[df_pt["value"] == patrol_segment.get("patrol_type")].reset_index()["id"][0]
+                patrol_type = (
+                    df_pt[df_pt["value"] == patrol_segment.get("patrol_type")].reset_index()["id"][0]
+                    if patrol_segment.get("patrol_type")
+                    else None
+                )
 
                 if None in {subject_id, patrol_start_time}:
                     continue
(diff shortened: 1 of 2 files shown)


cisaacstern (commented):
Question:

If data with some rows for which patrol type is `None` are sent through to ecoscope-workflows, and used in a Patrols workflow using a categorical/value grouper on `Patrol Type`...

does that imply that the following proposed / in-flight check in the `split_groups` will be at risk of erroring in this case? https://github.com/wildlife-dynamics/ecoscope-workflows/pull/698#discussion_r1964368416

For example:

```python
import pandas as pd

df = pd.DataFrame({'patrol_type': ['Ranger', 'Ranger']})
pd.api.types.is_string_dtype(df.patrol_type)
# --> True

df = pd.DataFrame({'patrol_type': ['Ranger', 'Ranger', None]})
pd.api.types.is_string_dtype(df.patrol_type)
# --> False
```

If so (and I think it does?), then we need to think about if we want to:

1. Filter out all rows for which a value grouper is `None` on the workflows side (so keep above-linked proposed check, but add a task to do this filtering)
2. Allow `None` in the Value Grouper as it's own category (in which case we want a check that permits `None`)

atmorling:
Based on the current (very different) thinking on the workflows value grouper PR - https://github.com/wildlife-dynamics/ecoscope-workflows/pull/698#discussion_r1965385278

IIUC, pandas `groupby` will drop `None` values by default, so I'm pretty sure we would get point 1. for free.
Point 2. is something I can add to the workflows PR as I think that's definitely going to be useful

walljcg:
Can we cast None to a string "None" so that we're not dropping anything before the groupby apply? This would be for any of the columns we want to use for value groupers.

atmorling:
@walljcg , makes sense. I will update the workflows PR to suit.

cisaacstern (approved):

PR #432, opened 2025-02-21
Fix LandDx settings and max_zoom

### LINKED ISSUE #433: LandDx map tile does not show up completely
The Mara Triangle monthly report now includes LandDx integration, but rendering issues persist. Additionally, when dealing with limited data, the map's automatic zoom level makes it difficult to view detailed information.

Yun-Wu:
Fixes #433


atmorling (approved):
LGTM! Thanks @Yun-Wu

PR #437, opened 2025-02-26
Optionally allow null event geometry, optionally 'coerce' non-point event geometry to points

### LINKED ISSUE #417: Allow get_events and get_patrol_events to optionally include null geometries
The task is to add a flag to optionally keep events where no is available.

Currently `filter_bad_geojson` will strip any and all rows where a location can't be parsed from the event geojson.

### LINKED ISSUE #436: Optionally convert polygon/multipolygon geometry to points in `get_events` and `get_patrol_events`
- We should default the option to `True`
- The converted point should be the centroid of the original poly

atmorling:
Closes #417
Closes #436


ecoscope/io/earthranger_utils.py
```diff
@@ -80,5 +81,22 @@ def pack_columns(dataframe: pd.DataFrame, columns: typing.List):
...
+    if df.empty:
+        return gpd.GeoDataFrame()
+
+    def shape_from_geojson(geojson):
+        try:
+            result = shape(geojson.get("geometry"))
+        except Exception:
+            return None
+
+        return result.centroid if force_point_geometry else result
```
cisaacstern, line 96: If the result is a Point then:

- `Point.centroid` exists as a property?
- And its an identity property? i.e. it's the same as the Point itself?

Just wanting to understand what happens if the result is a point, since I think _all_ geometry now passes through this check?

ecoscope/io/earthranger_utils.py
```diff
@@ -80,5 +81,22 @@ def pack_columns(dataframe: pd.DataFrame, columns: typing.List):
...
+    df: pd.DataFrame, geojson_column="geojson", force_point_geometry=True, drop_null_geometry=True
+):
+    if df.empty:
+        return gpd.GeoDataFrame()
+
+    def shape_from_geojson(geojson):
+        try:
+            result = shape(geojson.get("geometry"))
+        except Exception:
+            return None
```
cisaacstern, line 94: Is there ever a case where the geometry is malformed and we actually want to fail hard here, so the user _knows_ it's malformed? Rather than just silently "nullifying" it?

cisaacstern (changes requested):
Thanks @atmorling.

Request changes is because I'd like some more clarity on why we are allowing geometry to be set to `None` for _any_ exception thrown by the `shape(...` parsing. My hunch is that this seems like a whole through which a lot of geometries could disappear into null-ness 🕳️ ?

I would expect we'd want the exception type to be more specific than the builtin base `Exception` if we're doing this, and also perhaps to offer the option to fail hard on failed parsing?

But I am open to being convinced otherwise.

ecoscope/io/earthranger_utils.py
atmorling: Yep that's exactly what happens :+1: 
Centroid of a point == point

Perhaps this could be documented better, but the mixed geometry test demonstrates this [here](https://github.com/wildlife-dynamics/ecoscope/blob/c4882fdd029f09eae633ae47399ce11ebb49bb55/tests/sample_data/io/get_patrol_events_with_poly_geojson.json#L264-L271) and [here](https://github.com/wildlife-dynamics/ecoscope/blob/c4882fdd029f09eae633ae47399ce11ebb49bb55/tests/test_earthranger_io.py#L458-L460)

ecoscope/io/earthranger_utils.py
atmorling: So this is how it is today: currently if the `geometry` is ever malformed/such that shapely couldn't parse it, we would fail [hard](https://github.com/wildlife-dynamics/ecoscope/blob/76e25c244e17a184d4f2c1535b940773265e3d9e/ecoscope/io/earthranger.py#L638) because the conversion isn't 'safe' against that Shapely exception.

We haven't seen this scenario yet across the beta cohort, which would suggest that it probably doesn't happen very often (probably due to _how_ events in ER are collected)..

That said, we _are_ nulling out today if the geojson is ever malformed such that we can't resolve a 'geometry' key, and to me this feels like a safe extension of that. So to me it's kind of weird that we're silently nulling if the geojson is bad, but failing hard only if the geometry portion of that same geojson is bad.
I think in general since this is all happening per _event_ it's safer not to fail hard if a single event has bad geojson...

cisaacstern (approved):
Thanks @atmorling your responses have allayed my concerns!

PR #449, opened 2025-03-12
Flag smart tests

### LINKED ISSUE #448: Flag SmartIO tests seperately to the main IO tests
Since the SMART endpoint is a test endpoint with no uptime guarantees let's flag them separately so they don't create noise in the main CI tests

atmorling:
Closes #448

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,7 +43,7 @@ jobs:
           SMART_USERNAME: ${{ secrets.SMART_USERNAME }}
           SMART_PASSWORD: ${{ secrets.SMART_PASSWORD }}
         run: |
-          pytest -v -r s -m "${{ inputs.io }} and not smart_io" --color=yes --cov=ecoscope --cov-append --cov-report=xml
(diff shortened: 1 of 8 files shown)


.github/workflows/tests.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           SMART_USERNAME: ${{ secrets.SMART_USERNAME }}
           SMART_PASSWORD: ${{ secrets.SMART_PASSWORD }}
         run: |
-          pytest -v -r s -m "${{ inputs.io }}" --color=yes --cov=ecoscope --cov-append --cov-report=xml
+          pytest -v -r s -m "${{ inputs.io }} and not smart_io" --color=yes --cov=ecoscope --cov-append --cov-report=xml
```
Yun-Wu, line 46: Does this mean `smart_io` test will never been run in github CI?

atmorling: Correct , my thinking was that it just creates noise if the SMART API is down a lot of the time.
I realise that isn't ideal if one is explicitly developing against that API - I'll update this so that they still run on PR's

Yun-Wu (approved):
Thank you for doing this Alex! Looks great!

PR #456, opened 2025-03-20
Add Time Window Pagination in SMART IO

### LINKED ISSUE #351: Support SMART query for more than one month
Currently SMART throws a 500 error if the response size is too big. One workaround is to slice the time filter to weekly or monthly and send requests in parallel. Oliver also mentioned they may support pagination in the future. Before working on this issue, we need to check with Oliver to see their plan

Yun-Wu:
Closes #351

### DIFF
--- a/tests/test_smart_io.py
+++ b/tests/test_smart_io.py
@@ -29,3 +29,17 @@ def test_smart_get_patrol_observations(smart_io):
     assert "geometry" in result
     assert "groupby_col" in result
     assert "fixtime" in result
+
+
+def test_smart_get_patrol_observations_paginates(smart_io):
+    result = smart_io.get_patrol_observations(
+        ca_uuid="735606d2-c34e-49c3-a45b-7496ca834e58",
+        language_uuid="13451893-86af-4ec0-beac-2b8e0c2482b5",
+        start="2021-12-01",
+        end="2021-12-31",
+    )
+
+    assert len(result) > 0
+    assert "geometry" in result
+    assert "groupby_col" in result
+    assert "fixtime" in result
--- a/tests/test_smart_io.py
+++ b/tests/test_smart_io.py
@@ -36,7 +36,7 @@ def test_smart_get_patrol_observations_paginates(smart_io):
         ca_uuid="735606d2-c34e-49c3-a45b-7496ca834e58",
         language_uuid="13451893-86af-4ec0-beac-2b8e0c2482b5",
         start="2021-12-01",
-        end="2021-12-31",
+        end="2021-12-07",
     )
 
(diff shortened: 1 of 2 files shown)


atmorling (changes requested):
Change LGTM

Requesting changes because I think a test that guaruntees we'll hit the pagination logic, something like the below, would be helpful.

```python
def test_smart_get_patrol_observations_paginates(smart_io):
    result = smart_io.get_patrol_observations(
        ca_uuid="735606d2-c34e-49c3-a45b-7496ca834e58",
        language_uuid="13451893-86af-4ec0-beac-2b8e0c2482b5",
        start="2021-12-01",
        end="2021-12-31",
    )

    assert len(result) > 0
    assert "geometry" in result
    assert "groupby_col" in result
    assert "fixtime" in result
```

Yun-Wu:
> Change LGTM
> 
> Requesting changes because I think a test that guaruntees we'll hit the pagination logic, something like the below, would be helpful.
> 
> ```python
> def test_smart_get_patrol_observations_paginates(smart_io):
>     result = smart_io.get_patrol_observations(
>         ca_uuid="735606d2-c34e-49c3-a45b-7496ca834e58",
>         language_uuid="13451893-86af-4ec0-beac-2b8e0c2482b5",
>         start="2021-12-01",
>         end="2021-12-31",
>     )
> 
>     assert len(result) > 0
>     assert "geometry" in result
>     assert "groupby_col" in result
>     assert "fixtime" in result
> ```

Done

tests/test_smart_io.py
```diff
@@ -29,3 +29,17 @@ def test_smart_get_patrol_observations(smart_io):
...
     assert "groupby_col" in result
     assert "fixtime" in result
+
+
+def test_smart_get_patrol_observations_paginates(smart_io):
+    result = smart_io.get_patrol_observations(
+        ca_uuid="735606d2-c34e-49c3-a45b-7496ca834e58",
+        language_uuid="13451893-86af-4ec0-beac-2b8e0c2482b5",
+        start="2021-12-01",
+        end="2021-12-07",
```
atmorling, line 39: Is this not a 6 day difference? Ie; we won't paginate this range.

atmorling (changes requested):
Sorry to be that guy - requesting changes because the date range in the pagination results in total_duration=6

tests/test_smart_io.py
Yun-Wu: I reduced the window size to be 2 days since it seems to be more stable. Sorry to play around with magic numbers here

atmorling (approved):
LGTM - Thanks @Yun-Wu

PR #458, opened 2025-03-26
Loop colormap when values exceed the available colors

### LINKED ISSUE #457: `apply_colormap` should cycle colors when the number of categories exceeds the number of colors
Rather than the current implementation where all 'extra' categories are mapped to the final color value
This should apply to user defined cmaps as well.

xref discussion [here](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/750)

atmorling:
Closes #457


Yun-Wu (approved):
LGTM!

PR #467, opened 2025-04-22
Wrap gdf in base classes

### LINKED ISSUE #118: Support Geopandas >= 0.14.3
`EcoDataFrame` wraps `GeoDataFrame`, and assumes that when we call Geopandas/Pandas methods on `EcoDataFrame` objects that an `EcoDataFrame` is returned.

~~0.14.3 breaks that assumption in a few places, with internal methods that are marked `@final` meaning we can't wrap them on our side.~~

~~Because of that, 'supporting' 0.14.3 will probably mean updating our notebooks/documentation and tests to capture scenarios where a user will have convert back to an EcoDataFrame after certain operations~~

> @cisaacstern _**Note**_: per offline sync, we will do this by wrapping GeoDataFrame, as described in #440

atmorling:
Closes #118


ecoscope/base/base.py
```diff
@@ -338,50 +274,54 @@ class Trajectory(EcoDataFrame):
...
-    def get_tortuosity(self):
+    @staticmethod
+    def get_tortuosity(gdf: gpd.GeoDataFrame):
         """
         Get tortuosity for dataframe defined as distance traveled divided by displacement between first and final
         points.
         """
 
-        return self["dist_meters"].sum() / self.get_displacement()
+        return gdf["dist_meters"].sum() / Trajectory.get_displacement(gdf)
```
cisaacstern, line 324: Why are these `@staticmethod`s when they could (I think?) be regular instance methods that operate on `self.gdf`?

environment.yml
```diff
@@ -6,9 +6,8 @@ dependencies:
   - pip
   - git
   - jupyterlab
-  - fiona<1.10.0
-  - geopandas<=0.14.2
-  - numpy<2
+  - geopandas
+  - numpy<2.1
```
cisaacstern, line 10: Why numpy `<2.1`?

publish/recipes/release/ecoscope.yaml
```diff
@@ -39,7 +38,7 @@ requirements:
     # - matplotlib  # disabling in favor of matplotlib-base which is smaller
     - matplotlib-base
     - networkx
-    - numba
+    - numpy<2.1
```
cisaacstern, line 41: Same question as above re: np version

pyproject.toml
```diff
@@ -24,9 +24,8 @@ dependencies = [
   "backoff",
   "earthengine-api",
   "earthranger-client @ git+https://github.com/PADAS/er-client@v1.3.3",
-  "fiona<1.10.0",
-  "geopandas<=0.14.2",
-  "numpy<2",
+  "geopandas",
+  "numpy<2.1",
```
cisaacstern, line 28: I wonder if we should just specify `geopandas>1` to be very clear of the intention here?

pyproject.toml
```diff
@@ -61,6 +60,7 @@ test = [
   "pytest-cov",
   "pytest-mock",
   "pytest-reportlog",
+  "fsspec",
```
cisaacstern, line 63: Why are we adding `fsspec`?

cisaacstern (changes requested):
Really awesome.

Request changes is just for some minimal inline questions re:

- Some staticmethods in base.py
- Geopandas + numpy versions in pyproject.toml, etc.

pyproject.toml
atmorling: Seems reasonable

ecoscope/base/base.py
atmorling: IIUC, no reason other than "it's always been that way"
I can make that change with this PR

environment.yml
atmorling: I meant to flag this sorry. Was hitting a compatibility issue with stale deps in the Lonboard fork.
I want to get cracking with dusting that off / upstreaming things and will ticket out a follow up on this for next sprint.

ecoscope/base/base.py
atmorling: Ok looking further at the implications of this, the @staticmethods' are used in groupby applys in context that may or may not be outside of the specific trajectory/reloc context/instance.

Based on that I'd advocate for leaving this as is for the purpose of this PR and revisiting this with a broader review of the API.

cisaacstern: @atmorling I am confused by this statement

> IIUC, no reason other than "it's always been that way"

Because the diff above seems to clearly show that this PR is _making_ these into staticmethods, when they were previously instance methods, i.e. I am seeing as part of this PR's diff ...

```diff
-    def get_tortuosity(self):
+    @staticmethod
+    def get_tortuosity(gdf: gpd.GeoDataFrame):
```

Am I missing something here? This looks to me quite clearly as though this has _not_ always been this way.

environment.yml
cisaacstern: cool

cisaacstern (changes requested):
Thanks for the notes re: numpy versions, etc. All that makes sense.

Just dropping one more review here so that my lingering question re: staticmethods doesn't get lost in the inline notes.

This is my one remaining question/concern: https://github.com/wildlife-dynamics/ecoscope/pull/467#discussion_r2059497134

ecoscope/base/base.py
atmorling: Wow you're right, I completely forgot that I added that :facepalm: 
It _has not_ always been that way but it probably _should_ have always been that way.


>may or may not be outside of the specific trajectory/reloc context/instance.

In a context like the below, seen [here](https://github.com/wildlife-dynamics/ecoscope/blob/2643f72f59b96742563f251502597ae4a7f4bdc2/tests/test_base.py#L75-L84)
`trajectory.gdf.groupby("groupby_col").apply(ecoscope.base.Trajectory.get_displacement)` 
operating on `self` wouldn't make sense. This test magically works on main because self 'quacks' like a GDF.
Here it would have to be `self.gdf` which wouldn't work in the `apply` context because the `groupby` is on the unwrapped GDF.

ecoscope/base/relocations.py
```diff
@@ -0,0 +1,215 @@
...
+
+        default_cols = ["groupby_col", "fixtime", "junk_status", "geometry"]
+        extra_cols = gdf.columns.difference(default_cols)
+        extra_cols = extra_cols[~extra_cols.str.startswith("extra__")]
+
+        assert gdf.columns.intersection("extra__" + extra_cols).empty, "Column names overlap with existing `extra`"
+
+        gdf.rename(columns=dict(zip(extra_cols, "extra__" + extra_cols)), inplace=True)
+
+        return Relocations(gdf=gdf)
```
cisaacstern, line 85:
```suggestion
        return cls(gdf=gdf)
```
Do we want people to subclass this? Probably not. Might it happen? Sure. So I think this is advisable. Cf. https://stackoverflow.com/questions/46128877/should-i-use-the-class-name-or-cls-param-to-construct-a-an-instance

ecoscope/base/trajectory.py
```diff
@@ -0,0 +1,402 @@
...
+
+        original_crs = relocs.gdf.crs
+        relocs.gdf.to_crs(4326, inplace=True)
+        relocs.gdf = (
+            relocs.gdf.groupby("groupby_col")[relocs.gdf.columns].apply(cls._create_multitraj).droplevel(level=0)
+        )
+        relocs.gdf.to_crs(original_crs, inplace=True)
+
+        relocs.gdf.sort_values("segment_start", inplace=True)
+        return Trajectory(gdf=relocs.gdf)
```
cisaacstern, line 74:
```suggestion
        return cls(gdf=relocs.gdf)
```

Same comment as for `Relocations` classmethod.

ecoscope/base/relocations.py
```diff
@@ -0,0 +1,215 @@
...
+class Relocations(EcoDataFrame):
+    """
+    Relocation is a model for a set of fixes from a given subject.
+    Because fixes are temporal, they can be ordered asc or desc. The additional_data dict can contain info
+    specific to the subject and relocations: name, type, region, sex etc. These values are applicable to all
+    fixes in the relocations array. If they vary, then they should be put into each fix's additional_data dict.
+    """
+
+    @classmethod
+    def from_gdf(cls, gdf, groupby_col=None, time_col="fixtime", uuid_col=None, **kwargs):
```
cisaacstern, line 28:
```suggestion
    def from_gdf(cls, gdf, groupby_col=None, time_col="fixtime", uuid_col=None, copy: bool = True):
```
Since this is the only kwarg that's used in this method, it seems much clearer to name it explicitly.

ecoscope/base/relocations.py
```diff
@@ -0,0 +1,215 @@
...
+        gdf : GeoDataFrame
+            Observations data
+        groupby_col : str, optional
+            Name of `gdf` column of identities to treat as separate individuals. Usually `subject_id`. Default is
+            treating the gdf as being of a single individual.
+        time_col : str, optional
+            Name of `gdf` column containing relocation times. Default is 'fixtime'.
+        uuid_col : str, optional
+            Name of `gdf` column of row identities. Used as index. Default is existing index.
+        """
```
cisaacstern, line 41:
```suggestion
        copy : bool, optional
            Whether or not to copy the `gdf`. Defaults to `True`.
        """
```

ecoscope/base/relocations.py
```diff
@@ -0,0 +1,215 @@
...
+
+        assert {"geometry", time_col}.issubset(gdf)
+
+        if kwargs.get("copy") is not False:
+            gdf = gdf.copy()
```
cisaacstern, line 46:
```suggestion
        if copy:
            gdf = gdf.copy()
```

ecoscope/base/trajectory.py
```diff
@@ -0,0 +1,402 @@
...
+
+class Trajectory(EcoDataFrame):
+    """
+    A trajectory represents a time-ordered collection of segments.
+    Currently only straight track segments exist.
+    It is based on an underlying relocs object that is the point representation
+    """
+
+    @classmethod
+    def from_relocations(cls, relocs: Relocations, *args, **kwargs):
```
cisaacstern, line 47:
```suggestion
    def from_relocations(cls, relocs: Relocations, copy: bool = True):
```
Same comment as above. I think we should try to avoid open-ended `**kwargs` wherever possible. I realize that this is just copy-and-paste from the prior implementation, so not a conscious choice here, but it felt too glaring, and the fix just too easy, to not propose.

ecoscope/base/trajectory.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        Create Trajectory class from Relocation dataframe.
+        Parameters
+        ----------
+        relocs: Relocations
+            A `Relocations` instance.
+        args
+        kwargs
```
cisaacstern, line 55:
```suggestion
        copy: bool, optional
            Whether or not to copy the `relocs.gdf` attribute on construction.
            Defaults to `True`.
```

ecoscope/base/trajectory.py
```diff
@@ -0,0 +1,402 @@
...
+        Returns
+        -------
+        Trajectory
+        """
+        assert isinstance(relocs, Relocations)
+        assert {"groupby_col", "fixtime", "geometry"}.issubset(relocs.gdf)
+
+        if kwargs.get("copy"):
```
cisaacstern, line 63:
```suggestion
        if copy:
```

ecoscope/base/trajectory.py
```diff
@@ -0,0 +1,402 @@
...
+        return Geod(ellps="WGS84").inv(start[0], start[1], end[0], end[1])[2]
+
+    @staticmethod
+    def get_tortuosity(gdf: gpd.GeoDataFrame):
+        """
+        Get tortuosity for dataframe defined as distance traveled divided by displacement between first and final
+        points.
+        """
+
+        return gdf["dist_meters"].sum() / get_displacement(gdf)
```
cisaacstern, line 94:
```suggestion
```

These were moved to top level so can be deleted, correct?

cisaacstern (changes requested):
@atmorling thanks so much this is looking _really_ good. Personally, I think the separation of base into these distinct modules aids readability tremendously.

I have just a few nits below and then I think/assume we can close the book on this. The one of greatest/primary importance is the deletion of the static methods in Trajectory which have now been moved to the module top-level.

Otherwise, there are a few borderline stylistic things that I threw in just because I noticed them as I was reading, feel free to take or leave.

Thanks for the hard work here, this is a super impressive and significant effort!

ecoscope/base/trajectory.py
atmorling: :facepalm:

atmorling:
@cisaacstern  have updated with everything discussed here + slack

cisaacstern (approved):
LGTM! Thanks so much for this major effort! Any accommodating the reviews!

PR #477, opened 2025-05-07
Fix gdf clean

### LINKED ISSUE #476: map layer creation functions clean null geometry after unpacking style columns
This is unsafe because if we do drop anything we've mismatched the length of the style kwargs 🤦

atmorling:
closes #476 

In the spirit of #467 I have also factored these functions out of the class


Yun-Wu (approved):
LGTM!

PR #486, opened 2025-05-16
Add grid_size_from_geographic_extent

### LINKED ISSUE #446: Implement ETD and Feature Density grid size auto-scaling
Low-level implementation of https://github.com/wildlife-dynamics/compose/issues/561

Per refinement sync w/ @walljcg:

- Each of these algorithms should except _either_ an explicit grid size, or otherwise use an algorithm of some sort to automatically choose the grid size, with the latter being the default
- Though the main _purpose_ of this is to avoid over-running RAM, Jake and I discussed that we agree the implementation of the auto-scaling algorithm, for the purposes of the platform, should not take memory into account explicitly, as this could lead to non-deterministic behavior (affecting, e.g., snapshot testing), depending on where the workflow is running (and the resources available in that context)
- As a first pass, we can simply fit a curve (or line) to return grid size as a function of geographic extent only, such that the values returned by that function align with the values we have seen to work thus far in our deployment, i.e.

  ```python
  def grid_size_from_geographic_extent(geographic_extent: tuple[float, float]) -> tuple[float, float]: ...
  
  global_extent = ...
  grid_size_from_geographic_extent(global_extent)  # --> 5,000
  
  regional_extent = ...
  grid_size_from_geographic_extent(regional_extent)  # --> 250
  
  ```

atmorling:
Closes #446


tests/test_utils.py
```diff
@@ -1,13 +1,14 @@
...
-from ecoscope.base.utils import hex_to_rgba
+
 from ecoscope.base.utils import (
     create_meshgrid,
     groupby_intervals,
     create_interval_index,
     create_modis_interval_index,
     add_val_index,
     add_temporal_index,
+    hex_to_rgba,
```
atmorling, line 11: Unrelated import cleanup

ecoscope/analysis/UD/etd_range.py
```diff
@@ -240,3 +242,26 @@ def calculate_etd_range(
...
+    and divides by the scale_factor to determine a 'sensible' grid size in meters
+    """
+    gdf = gdf.to_crs("EPSG:4326")
+
+    def _haversine_diagonal(bbox: BoundingBox):
+        # From https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html
+        bbox_radians = [math.radians(x) for x in bbox]
+        min_point = [bbox_radians[0], bbox_radians[1]]
+        max_point = [bbox_radians[2], bbox_radians[3]]
+        result = haversine_distances([min_point, max_point])
```
walljcg, line 260: For the sake of consistency we should use the same distance calculation we use for the trajectory calculations. cf line 28 of trajectory.py:

from pyproj import Geod
Geod(ellps="WGS84").inv(start[0], start[1], end[0], end[1])[2]

This already calculates in meters.

walljcg (changes requested):

ecoscope/analysis/UD/etd_range.py
atmorling: Somehow completely missed we were doing that in `get_displacement` :exploding_head: Will do!

PR #491, opened 2025-05-26
Remove barmode in bar chart

### LINKED ISSUE #492: Remove barmode in bar chart
So that it can be configurable from the workflows

Yun-Wu:
Closes #492


atmorling (approved):
LGTM

PR #501, opened 2025-07-18
Add linear time density function

### LINKED ISSUE #494: Add a linear time density analysis to core library
- [ ] TBD with @walljcg whether we'll use scipy or hand-roll our own implementation.

- [ ] Verify with @walljcg that the new snapshot looks correct

atmorling:
Closes #494 

- Adds a linear time density calculation function
- Adds a generalised version of `get_percentile_area` to do percentile classification over a dataframe column

### DIFF
--- a/ecoscope/analysis/linear_time_density.py
+++ b/ecoscope/analysis/linear_time_density.py
@@ -23,8 +23,7 @@ def calculate_ltd(
     grid = grid.reset_index().rename(columns={"index": "grid_cell_index"})
     overlay = grid.overlay(traj.gdf, how="intersection", keep_geom_type=False)
 
-    # drop anything that isn't a line but keep the total timespan
-    dropped_time = max(overlay[overlay.geometry.type != "LineString"]["timespan_seconds"].sum(), 0)
+    # drop anything that isn't a line
     overlay = overlay[overlay.geometry.type == "LineString"]
 
     overlay["fragment_distance"] = overlay["geometry"].apply(
@@ -39,10 +38,6 @@ def calculate_ltd(
     grid["density"] = overlay.groupby("grid_cell_index")["fragment_time"].sum()
 
     total_time = round(grid["density"].sum(), 1)
-
-    assert (
-        total_time == traj.gdf["timespan_seconds"].sum() - dropped_time
-    ), "Segment times do not sum to the Trajectory's total time, this is likely an issue with the provided mesh grid"
 
     grid["density"] = grid["density"] / total_time
     return grid


atmorling:
This is ready for review but I'll add some more test coverage for `classify_percentile` prior to merge

ecoscope/analysis/classifier.py
```diff
@@ -164,3 +165,46 @@ def apply_color_map(
...
+    Returns:
+    The input dataframe with percentile classification appended.
+    """
+    assert pd.api.types.is_numeric_dtype(gdf[input_column_name]), "input column must contain numeric values"
+
+    input_values = gdf[input_column_name].to_numpy()
+    input_values = np.sort(input_values[~np.isnan(input_values)])
+    csum = np.cumsum(input_values)
+
+    percentile_values = []
```
walljcg, line 197: Is the numpy percentiles function a good option here? https://numpy.org/doc/2.2/reference/generated/numpy.percentile.html

ecoscope/analysis/linear_time_density.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Returns
+    -------
+    grid : gpd.GeoDatFrame
+    """
+
+    # We need the grid cell index later to count density values
+    grid = grid.reset_index().rename(columns={"index": "grid_cell_index"})
+
+    overlay = grid.overlay(traj.gdf, how="intersection", keep_geom_type=False)
```
walljcg, line 26: May consider keeping only line types after this operation in case ever have the start/end of a line segment exactly on the boundary of a grid cell which would likely result in a point geometry being created.

ecoscope/analysis/linear_time_density.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # fragment_distance is in metres so convert speed to 'meters per hour'
+    # then multiply by 3600 to get time in seconds
+    overlay["fragment_time"] = (overlay["fragment_distance"] / (overlay["speed_kmhr"] * 1000)) * 3600
+
+    # This is our time density value
+    grid["density"] = overlay.groupby("grid_cell_index")["fragment_time"].sum()
+
+    total_time = round(grid["density"].sum(), 1)
+    assert total_time == traj.gdf["timespan_seconds"].sum()
```
walljcg, line 39: Are we rounding to a single digit on timespan_seconds as well?  Impressed this assertion works given rounding errors!

ecoscope/analysis/linear_time_density.py
```diff
@@ -0,0 +1,42 @@
...
+    grid : gpd.GeoDatFrame
+    """
+
+    # We need the grid cell index later to count density values
+    grid = grid.reset_index().rename(columns={"index": "grid_cell_index"})
+
+    overlay = grid.overlay(traj.gdf, how="intersection", keep_geom_type=False)
+    overlay["fragment_distance"] = overlay["geometry"].apply(
+        lambda x: Geod(ellps="WGS84").inv(*x.coords[0], *x.coords[1])[2]
+    )  # Geod.inv returns tuple where the 3rd value is the distance in metres
```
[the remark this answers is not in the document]
walljcg: Nice!

walljcg (commented):
Looks good to me! Nice work

ecoscope/analysis/linear_time_density.py
atmorling: So double checking this...`timespan_seconds` is actually rounded to `.0`, so the fact that `round(grid["density"].sum(), 1)` also results in `##.0`, which is pretty cool.

I'll update this to give a descriptive error if it ever fails. 
Perhaps we can handle this at the workflows level by showing no data for the time density map, rather than killing the whole workflow

ecoscope/analysis/linear_time_density.py
atmorling: Will do!

ecoscope/analysis/classifier.py
atmorling: I tested `np.percentile` and `pd.quantile` and observed that np.percentile generates values that tend to be smaller than the calculation here. `pd.quantile` is much closer to the cumsum approach, but ultimately opted to stick with what's used elsewhere in this library.

ecoscope/analysis/linear_time_density.py
cisaacstern: +1 on not killing the whole workflow 😄

cisaacstern (approved):
Full transparency, the Science™️ is a little over my head, but LGTM from a code standpoint, thanks!

ecoscope/analysis/linear_time_density.py
atmorling: Realising that we actually should drop this assertion if we're going to implement the above suggestion about dropping any non-line geometries generated by the overlay.

If we ever drop data from the overlay, grid["density"] will no longer sum to the sum of the original `timespan_seconds`

PR #504, opened 2025-07-24
Use EPSG:3857 in `create_meshgrid`, instead of `estimate_utm_crs`

### LINKED ISSUE #489: Default to using EPSG:3857 in create_meshgrid
Datasets that cover a large area will experience distortion when plugged into `create_meshgrid` due to our use of `estimate_utm_crs`. We should by default use EPSG:3857 as a more general-case alternative.

atmorling:
Closes #489 

- `create_meshgrid` will now reproject the input AOI to the `out_crs` to calculate the meshgrid cells (instead of estimating a utrm crs)
- Removed default xlen/ylen values from `grid_to_raster`, since we should be agnostic of CRS here
- `calculate_feature_density` and `calculate_linear_time_density` not project the input gdf to the CRS of the provided meshgrid.
  - `calculate_linear_time_density` will throw an error if the provided meshgrid is a projected but not in metres


atmorling:
@cisaacstern @walljcg sorry for the noise - this is ready for review (again)

cisaacstern (approved):
With the caveat that I am very much not up-to-speed on this area of code, from a code standpoint LGTM! Thanks Alex!

PR #516, opened 2025-08-26
Allow setting page size per request

### LINKED ISSUE #515: Allow setting sub_page_size per request
For 
`get_events`
`get_patrol_observations_with_patrol_filter`
`get_patrol_events` 
`get_subjectgroup_observations` 

we want to add a page size param that would override the client default if present

atmorling:
Closes #515 

Allows overriding the client default `sub_page_size` param passed to `get_objects_multithreaded`

### DIFF
--- a/ecoscope/base/utils.py
+++ b/ecoscope/base/utils.py
@@ -102,10 +102,10 @@ def groupby_intervals(df: pd.DataFrame, col: str, intervals: pd.IntervalIndex) -
             interval: df.iloc[left_i:right_i]
             for interval, left_i, right_i in zip(
                 intervals,
-                df[col].searchsorted(intervals.left, side="left" if intervals.closed in ("left", "both") else "right"),
+                df[col].searchsorted(intervals.left, side="left" if intervals.closed in ("left", "both") else "right"),  # type: ignore[call-overload]
                 df[col].searchsorted(
                     intervals.right, side="right" if intervals.closed in ("right", "both") else "left"
-                ),
+                ),  # type: ignore[call-overload]
             )
         }
     ).groupby(level=0)


tests/test_earthranger_io.py
```diff
@@ -477,3 +478,137 @@ def test_empty_responses(_get_mock, er_io, er_callable, er_kwargs):
...
+@patch("ecoscope.io.earthranger.EarthRangerIO.get_objects_multithreaded")
+def test_page_size_override_with_patrols(get_objects_mock, set_sub_page_size, er_callable, er_kwargs, er_io):
+    """
+    The intent of this test is to make sure that if page size is set on the er_callable,
+    _all_ downstream calls to get_objects_multithreaded use the overridden page_size
+    """
+    patrols_json = json.loads(open("tests/sample_data/io/sample_patrol.json", "r").read())
+
+    def patrols_side_effect(*args, **kwargs):
+        yield patrols_json if get_objects_mock.call_count == 1 else MagicMock()
```
atmorling, line 521: It's necessary to mock this response so that we avoid hitting empty df logic which would skip the downstream calls to `get_objects_multithreaded` that we're trying to test

tests/test_earthranger_io.py
```diff
@@ -477,3 +478,137 @@ def test_empty_responses(_get_mock, er_io, er_callable, er_kwargs):
...
+    """
+    The intent of this test is to make sure that if page size is set on the er_callable,
+    _all_ downstream calls to get_objects_multithreaded use the overridden page_size
+    """
+    subjects_json = json.loads(open("tests/sample_data/io/sample_subject.json", "r").read())
+    observation_json = json.loads(open("tests/sample_data/io/sample_observation.json", "r").read())
+
+    def patrols_side_effect(*args, **kwargs):
+        yield subjects_json
+        yield observation_json
```
atmorling, line 552: Same as above comment

tests/conftest.py
atmorling, line 1: Unrelated changes, removing an unnecessary fixture marks.

PR #542, opened 2025-10-21
Reconcile event display names

### LINKED ISSUE #541: Allow reconciling event type display names in events dataframes
Currently `get_events` only returns the event_type value, if we want to include the display names of the event types we'll need to query `get_event_types` in order to reconcile the values to display names.

It would also be nice if we could include the event category display name in the cases where the event type display names overlap

atmorling:
## :earth_americas: Summary
Closes #541 

## :package: Proposed Changes
- Adds a `get_event_type_display_names_from_events` function to EarthrRangerIO to resolve event type display values for a given dataframe of events
  - This function also optionally appends the relevant event category  to the resolved display name to handle ER sites that use overlapping display names for different event types/categories

### DIFF
--- a/ecoscope/io/earthranger.py
+++ b/ecoscope/io/earthranger.py
@@ -793,15 +793,17 @@ class EarthRangerIO(ERClient):
         if not do_append:
             return events_gdf
 
-        event_categories = self.get_event_categories()
-        event_categories_lookup = {category["value"]: category["display"] for category in event_categories}
-        event_type_to_category_display_lookup = {
-            event_type["value"]: event_categories_lookup[
-                # In V1 event types the category information is a dict, with V2 event types category == category value
-                event_type["category"]["value"] if isinstance(event_type["category"], dict) else event_type["category"]
-            ]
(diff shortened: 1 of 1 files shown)


ecoscope/io/earthranger.py
```diff
@@ -756,6 +757,68 @@ def get_events(
...
+            category["value"]: category["display"] for category in self.get_event_categories()
+        }
+        # In V1 event types, event category information is bundled as a nested dict
+        # In V2 event types we only get the event category value at the same "category" key
+        # So we need to safely handle that difference here
+        categories_to_lookup = event_types["category"].apply(
+            lambda category: category["value"] if isinstance(category, dict) else category
+        )
+        category_display_values = categories_to_lookup.map(event_categories_display_lookup)
+        event_type_to_category_display_lookup = dict(zip(event_types["value"], category_display_values))
```
atmorling: The end result here is a mapping of event_type_value to event_category_display_name, but it's kind of a roundabout way of resolving this due to having to reconcile information between a datframe of events and a dict of event categories.

PR #555, opened 2025-12-24
Allow `calculate_feature_density` to sum a numeric column (instead of counting rows)  

### LINKED ISSUE #554: Add counting column values to `calculate_feature_density`
Optionally allow summing the values of a numeric column in the provided dataframe.

atmorling:
## :earth_americas: Summary
Closes #554

## :package: Proposed Changes
- Adds a `sum_column` param such that, when provided, rather than counting the geometry that falls within the clip, will sum the `sum_column` for geometry that falls within the clip.
- Fixes a flakey IO test by reducing the amount of data queried

### DIFF
--- a/tests/test_feature_density.py
+++ b/tests/test_feature_density.py
@@ -13,7 +13,7 @@ def test_feature_density_point():
 
     grid = gpd.GeoDataFrame(
         geometry=ecoscope.base.utils.create_meshgrid(
-            AOI.unary_union, in_crs=AOI.crs, out_crs="EPSG:3857", xlen=5000, ylen=5000, return_intersecting_only=False
+            AOI.union_all(), in_crs=AOI.crs, out_crs="EPSG:3857", xlen=5000, ylen=5000, return_intersecting_only=False
         )
     )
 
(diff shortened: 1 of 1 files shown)


ecoscope/analysis/feature_density.py
```diff
@@ -2,19 +2,30 @@
...
 import numpy as np
+from pandas.api.types import is_numeric_dtype
 
 
 def calculate_feature_density(
-    selection: gpd.GeoDataFrame, grid: gpd.GeoDataFrame, geometry_type: Literal["point", "line"] = "point"
+    selection: gpd.GeoDataFrame,
+    grid: gpd.GeoDataFrame,
+    geometry_type: Literal["point", "line"] = "point",
+    sum_column: str | None = None,
```
Yun-Wu: What do you think if we will use other aggregation functions in feature density calculation besides `sum`?

atmorling: Love it. I'll update soon!

Yun-Wu (approved):

PR #586, opened 2026-02-09
Support smoothing and apply it to line_chart

### LINKED ISSUE #587: Support smoothing and apply it to line_chart
This ticket is to solve the issue when interpolating precipitation when the value should never be negative

Yun-Wu:
## :earth_americas: Summary
Add smoothing and apply it to line_chart. Closes #587 

## :package: Proposed Changes
Added a new SmoothingConfig that allows configurable spline smoothing with value clamping. When smoothing is enabled, the chart renders two layers:

1. A smoothed spline line (with hover disabled)
2. Original data point markers (for accurate hover values)

### DIFF
--- a/ecoscope/analysis/smoothing.py
+++ b/ecoscope/analysis/smoothing.py
@@ -67,7 +67,7 @@ def apply_smoothing(x: np.ndarray, y: np.ndarray, config: SmoothingConfig) -> tu
 
     # Check for datetime types (numpy datetime64 or Python date/datetime objects)
     is_numpy_datetime = np.issubdtype(x_sorted.dtype, "datetime64")
-    is_python_date = len(x_sorted) > 0 and isinstance(x_sorted[0], (datetime.date, datetime.datetime))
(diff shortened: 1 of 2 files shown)


ecoscope/analysis/smoothing.py
```diff
@@ -0,0 +1,97 @@
...
+        The smoothing configuration
+
+    Returns
+    -------
+    tuple[np.ndarray, np.ndarray]
+        The smoothed (x, y) values
+    """
+
+    if config.method != "spline":
+        raise ValueError(f"Unsupported smoothing method: {config.method}")
```
atmorling:
```suggestion
        raise NotImplementedError(f"Unsupported smoothing method: {config.method}")
```

Minor nit but this should raise `NotImplementedError`
Maybe the better option here is to remove the `method` param entirely for now while we only support `spline`

atmorling (changes requested):
LGTM overall, have requested changes on account of a few implementation questions I have, mostly [this](https://github.com/wildlife-dynamics/ecoscope/pull/586/changes#r2781820222) one

PR #591, opened 2026-02-10
Resolve SMART API uuids from CA / Language names

### LINKED ISSUE #589: Resolve SMART CA UUID on init from either the CA ID or Name
Rather than requiring users to know the UUID, we should allow the option of initing from the CA ID or Name

atmorling:
## :earth_americas: Summary
Closes #589 

## :package: Proposed Changes
- Fleshes out the init handling of the SMART client allowing user's to provide either the CA ID or CA Name and optionally a Language Code.
- Removes `ca_uuid` and `language_uuid` args from user facing `get_` functions, these are now handled by the client.

### DIFF
--- a/ecoscope/io/smartio.py
+++ b/ecoscope/io/smartio.py
@@ -83,7 +83,7 @@ class SmartIO:
                 raise ValueError(f"Conservation area name '{self._ca_name}' not found in API response")
             return matches.iloc[0]["uuid"]
 
-        raise ValueError("No conservation area identifier provided. Must provide ca_uuid, ca_id, or ca_name")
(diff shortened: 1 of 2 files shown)


ecoscope/io/smartio.py
```diff
@@ -14,16 +14,34 @@
...
+    def __init__(
+        self,
+        urlBase: str,
+        username: str,
+        password: str,
+        ca_uuid: str = "",
+        ca_id: str = "",
+        ca_name: str = "",
+        language_uuid: str = "",
+        language_code: str = "en",
```
Yun-Wu: Not sure if we want to make `ca_uuid` and `language_uuid` part of the SMART client as some workflows may want to fetch data from multiple CAs

PR #598, opened 2026-02-12
Allow looking up a spatial feature group from the group name

### LINKED ISSUE #597: Allow group name as input to `get_spatial_features_group`
If a group name, rather than the ID is entered we can lookup all groups and resolve to the id. From there we can lookup the group as normal.

atmorling:
## :earth_americas: Summary
Closes #597 

## :package: Proposed Changes
- If group name is provided, we'll first lookup all spatial feature groups to resolve the UUID
- Then lookup the full geometry from the UUID as normal
- We'll raise if the provided name isn't resolvable to any UUID

### DIFF
--- a/ecoscope/io/earthranger.py
+++ b/ecoscope/io/earthranger.py
@@ -1221,7 +1221,11 @@ class EarthRangerIO(ERClient):
         return df
 
     def get_spatial_features_group(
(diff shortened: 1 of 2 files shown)


cisaacstern (approved):
LGTM

PR #627, opened 2026-03-20
ER Warehouse Client Integration

### LINKED ISSUE #628: ER Warehouse Client Integration
As the Ecoscope team
We want to integrate the warehouse client into selected workflow tasks (conditionally on an env var)
So that they retrieve observations from the warehouse instead of the ER API

marianobrc:
## :earth_americas: Summary
closes #628

Replaces the ER API client with the ER Warehouse client in workflow tasks reading subject observations or patrol observations (events are out of scope).

## :package: Proposed Changes
- Adds ecoscope-earthranger-io-core as a dependency to get `ERWarehouseClient`
- Updates get_subjectgroup_observations, get_patrol_observations, and get_patrol_observations_from_patrols_df tasks to use the ERWarehouseClient client, conditionally on env var USE_EARTHRANGER_WAREHOUSE_API
- Adds .server and .token attributes in  `EarthRangerClientProtocol` as [agreed in the previous PR review](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/1315/changes#r2858335482)
- Adds comments about features which are not supported in the warehouse yet: exclusion flags filter and `patrols_overlap_daterange=False` (default warehouse behavior matches `patrols_overlap_daterange=True`)

## Checklist
- [X] Unit tests updated if necessary

### DIFF
--- a/publish/recipes/release/ecoscope.yaml
+++ b/publish/recipes/release/ecoscope.yaml
@@ -76,7 +76,6 @@ outputs:
         - pydantic <2.9.0
         - pydantic-settings
         - cloudpathlib-gs
-        - astropy
         - wt-registry >=0.2.0,<0.3.0
         - wt-task >=0.1.0,<0.2.0
     tests:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,6 @@ platform = [
     "pydantic<2.9.0",
     "pydantic-settings",
     "cloudpathlib[gs]",
-    "astropy",
     "wt-registry>=0.2.0,<0.3.0",
     "wt-task>=0.1.0,<0.2.0",
     "ecoscope-earthranger-io-core @ git+https://github.com/wildlife-dynamics/ecoscope-earthranger-io-core.git@v0.0.4",


ecoscope/io/earthranger.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, sub_page_size: int = 4000, tcp_limit: int = 5, **kwargs):
             kwargs["service_root"] = f"{self.server}/api/v1.0"
             kwargs["token_url"] = f"{self.server}/oauth2/token"
 
+        self.token = kwargs.get("token")
```
cisaacstern:
```suggestion
```

`EarthRangerIO` inherits from `ERClient` which already sets `.token` if `"token"` is in kwargs; xref https://github.com/PADAS/er-client/blob/71ffa736d235e89178dd7a316221e6ed50da0bec/erclient/client.py#L97

So this line is duplicative of that one, and can be dropped I think. Did you add this for purely logical reasons or was the type checker complaining? Perhaps we can do an `if TYPE_CHECKING` block (or similar) if the latter.

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -20,6 +22,28 @@
...
+def _make_warehouse_client_from_env(er_site_url: str, er_api_token):
+    """Create an ERWarehouseClient if warehouse env vars are configured.
+
+    Returns None when the warehouse is not enabled, causing callers to
+    fall back to the legacy EarthRanger API client.
+    """
+    if os.environ.get("USE_EARTHRANGER_WAREHOUSE_API", "false").lower() == "true" and (
+        warehouse_api_base_url := os.environ.get("EARTHRANGER_WAREHOUSE_API_BASE_URL")
+    ):
+        from ecoscope_earthranger_io_core.client import ERWarehouseClient  # type: ignore[import-untyped]
```
cisaacstern: Nit: it _should_ be fine to move this import to the top level, since everyting in the `.platform` namespace requires the `ecoscope[platform]` extra, so if we try to import this module, `ecoscope_earthranger_io_core` deps should be available.

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -20,6 +22,28 @@
...
+
+def _make_warehouse_client_from_env(er_site_url: str, er_api_token):
+    """Create an ERWarehouseClient if warehouse env vars are configured.
+
+    Returns None when the warehouse is not enabled, causing callers to
+    fall back to the legacy EarthRanger API client.
+    """
+    if os.environ.get("USE_EARTHRANGER_WAREHOUSE_API", "false").lower() == "true" and (
+        warehouse_api_base_url := os.environ.get("EARTHRANGER_WAREHOUSE_API_BASE_URL")
+    ):
```
cisaacstern:
```suggestion
    ) and (
        er_api_token is not None
    ):
```

Let's guard against the variant where er_api_token is None here, by avoiding ERWarehouseClient construction entirely, since even though the client will technically allow itself to be constructed without a token, it will error [on network call](https://github.com/wildlife-dynamics/ecoscope-earthranger-io-core/blob/776ebf26d9add101ca80bfa85189a2c726d3774d/src/ecoscope_earthranger_io_core/client.py#L102-L105) when trying to construct auth headers.

... I suppose an alternative to adding the `er_api_token in not None` check here would be to do the same check inside this `if:` block and fail hard with a ValueError + descriptive message; I'm open to that option too if you prefer it.

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -20,6 +22,28 @@
...
+        from ecoscope_earthranger_io_core.client import ERWarehouseClient  # type: ignore[import-untyped]
+
+        logger.debug("Using ERWarehouseClient with base_url=%s", warehouse_api_base_url)
+        return ERWarehouseClient(
+            server=er_site_url,
+            token=er_api_token,
```
cisaacstern: Note that in changing the type of `er_api_token` to SecretStr as noted above we do _not_ have to call `.get_secret_value()` here (i.e. we can leave this as-is, because ERWarehouseClient handles that internally itself.

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -408,19 +432,37 @@ def get_subjectgroup_observations(
...
+
+        table = warehouse_client.get_subjectgroup_observations(
+            subject_group_name=subject_group_name,
+            include_subject_details=True,
+            include_inactive=True,
+            include_details=include_details,
+            include_subjectsource_details=include_subjectsource_details,
+            since=time_range.since.isoformat(),
+            until=time_range.until.isoformat(),
+            # ToDo: Pass exclusion flags filter once supported in the DWH API
```
cisaacstern: Depending on how long you think it may be before we have this feature, I wonder if we should put in a `logger.warning` here re: the fact that filter is not supported yet? Or if that will be a fast follow perhaps unecessary.

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -445,17 +487,35 @@ def get_patrol_observations(
...
-        status=status,
-        include_patrol_details=include_patrol_details,
-        sub_page_size=sub_page_size,
-        patrols_overlap_daterange=patrols_overlap_daterange,
-    )
-    if isinstance(patrol_obs_relocs, Relocations):
-        patrol_obs_relocs = patrol_obs_relocs.gdf
+    if warehouse_client := _make_warehouse_client_from_env(
+        er_site_url=client.server,
+        er_api_token=getattr(client, "token", None),
```
cisaacstern:
```suggestion
        er_api_token=SecretStr(client.token) if hasattr(client, "token") else None,
```

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -653,13 +713,28 @@ def get_patrol_observations_from_patrols_df(
...
-    patrol_obs_relocs = client.get_patrol_observations(
-        patrols_df=patrols_df,
-        include_patrol_details=include_patrol_details,
-        sub_page_size=sub_page_size,
-    )
-    if isinstance(patrol_obs_relocs, Relocations):
-        patrol_obs_relocs = patrol_obs_relocs.gdf
+    if warehouse_client := _make_warehouse_client_from_env(
+        er_site_url=client.server,
+        er_api_token=getattr(client, "token", None),
```
cisaacstern:
```suggestion
        er_api_token=SecretStr(client.token) if hasattr(client, "token") else None,
```

ecoscope/platform/connections.py
```diff
@@ -68,6 +68,9 @@ def connection_from_client(obj) -> DataConnection:
 
 @runtime_checkable
 class EarthRangerClientProtocol(Protocol):
+    server: str
+    token: str | None
```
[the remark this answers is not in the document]
cisaacstern: As a general (non-blocking) question, do you think it would be helpful to move this Protocol into `ecoscope-earthranger-io-core` ?

PR #666, opened 2026-04-15
Add auto_detect param to convert_values_to_timezone and doc update

### LINKED ISSUE #665: Add auto_detect param to convert_values_to_timezone
## Summary

Add an `auto_detect` boolean parameter to the `convert_values_to_timezone` task. When `True`, auto-detect all timezone-aware datetime columns (`DatetimeTZDtype`) instead of requiring an explicit column list.

## Motivation

In workflows like `wt-download-events`, event detail columns are normalized from JSON and have dynamically-generated column names that aren't known at spec-time. The current task requires explicitly listing columns, which can't cover dynamic datetime columns. With `auto_detect=True`, a second invocation of this task can convert all remaining datetime columns after normalization.

## Changes

- Add `auto_detect: bool = False` parameter to `convert_values_to_timezone`
- When `True`, discover all `DatetimeTZDtype` columns and convert them, ignoring the `columns` list
- Add test for the new behavior

Yun-Wu:
## Summary

- Add `auto_detect: bool = False` parameter to `convert_values_to_timezone` task
- When `True`, auto-detect all `DatetimeTZDtype` columns instead of requiring an explicit column list
- Enables converting dynamically-named event detail columns after JSON normalization in workflows
- Update docs on handling this package in editable mode

## Test plan

- [x] Added `test_convert_values_to_timezone_auto_detect` — verifies that all datetime columns are detected and converted, and non-datetime columns are left untouched
- [x] Existing parametrized tests still pass (CI)

Closes #665

### DIFF
--- a/doc/platform-sdk/content/troubleshooting.md
+++ b/doc/platform-sdk/content/troubleshooting.md
@@ -73,6 +73,29 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
 
 ---
 
+## "Could not read properties of undefined (reading 'name')"
+
+This error appears in Ecoscope Desktop / Ecoscope Web when submitting the configuration form. It means your workflow is missing one or both of the required task instances: `workflow_details` and `time_range`.
+
+Every ecoscope workflow **must** include both:
+
+```yaml
+workflow:
+  - name: Workflow Details
+    id: workflow_details
+    task: set_workflow_details
+
+  - name: Time Range
+    id: time_range
+    task: set_time_range
+    partial:
+      time_format: '%d %b %Y %H:%M:%S'
+```
+
+**Fix**: Add the missing task(s) to your `spec.yaml`. The `id` values must be exactly `workflow_details` and `time_range` — Ecoscope Desktop and Ecoscope Web look for these specific IDs when processing the configuration form.
+
+---
+
 ## Developing against a local ecoscope checkout
 
 When developing or debugging ecoscope itself, you can point your workflow at a local checkout instead of the published `ecoscope-platform` conda package. The conda package name is `ecoscope-platform`, but the PyPI/local package name is `ecoscope` with extras:
@@ -83,6 +106,9 @@ requirements:
     version: "3.12.*"
   - name: rasterio
     version: "1.5.0"
+    channel: conda-forge
+  - name: numpy
+    version: ">=2,<2.1"
     channel: conda-forge
   - name: ecoscope
     path: /Users/me/ecoscope
@@ -97,23 +123,35 @@ requirements:
   pre-built wheels.
 - **`rasterio` conda dep** — ensures GDAL native libraries are available for
   the editable ecoscope install.
-- **`extras`** — all three are required for task discovery to succeed:
-    - `platform`: pandera, pydantic, wt-registry, wt-task
+- **`extras`** — the extras you need depend on which tasks your workflow uses and the dependencies.
+  The four shown below cover the most common case:
+    - `platform`: pandera, pydantic, wt-registry, wt-task (always required)
     - `mapping`: lonboard, matplotlib (results/map tasks)
     - `analysis`: statsmodels (analysis tasks)
-- **Post-compile numpy patch** — ecoscope currently pins `numpy>=2,<2.1`,
-  which conflicts with what conda resolves on some platforms. After compiling,
-  manually edit the generated `pixi.toml` to pin `numpy>=2,<2.1` to match.
-  This constraint will be relaxed in a future ecoscope release.
+    - `plotting`: plotly, scikit-learn
+
+    Other extras may also be needed depending on your tasks/dependencies.
+- **`numpy` pin** — ecoscope currently pins `numpy>=2,<2.1`, which conflicts
+  with what conda resolves on some platforms. Add a numpy conda requirement to
+  the spec to avoid a post-compile patch:
+
+    ```yaml
+    - name: numpy
+      version: ">=2,<2.1"
+      channel: conda-forge
+    ```
+
 - **Post-compile pydantic patch** — ecoscope requires `pydantic<2.9.0` (newer
   versions break discriminated unions in `apply_classification`), but the
   compiler emits `pydantic>=2.0.0,<3.0.0` as a conda dependency. After
   compiling, manually edit the generated `pixi.toml` to pin
-  `pydantic>=2.0.0,<2.9.0`.
-- **Separate environment limitation** — `wt-compiler` requires `pydantic>=2.9` while
+  `pydantic>=2.0.0,<2.9.0`. This cannot be set via `requirements:` because
+  the compiler overrides pydantic's version constraint.
+- **Separate environments** — `wt-compiler` requires `pydantic>=2.9` while
   `ecoscope[platform]` requires `pydantic<2.9.0`. They cannot coexist in the
-  same pixi environment. The compiler does not import ecoscope at runtime — it
-  installs it in an ephemeral subprocess environment for task discovery.
+  same pixi environment, which is why the compiler installs requirements in an
+  ephemeral subprocess environment for task discovery rather than importing
+  them directly.
 
 ---
 
--- a/doc/platform-sdk/content/troubleshooting.md
+++ b/doc/platform-sdk/content/troubleshooting.md
@@ -75,21 +75,15 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
 
 ## "Could not read properties of undefined (reading 'name')"
 
-This error appears in Ecoscope Desktop / Ecoscope Web when submitting the configuration form. It means your workflow is missing one or both of the required task instances: `workflow_details` and `time_range`.
+This error appears in Ecoscope Desktop / Ecoscope Web when submitting the configuration form. It means your workflow is missing the required task instance `workflow_details`.
 
-Every ecoscope workflow **must** include both:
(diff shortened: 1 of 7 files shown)


doc/platform-sdk/content/troubleshooting.md
```diff
@@ -73,6 +73,50 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
...
+    - `analysis`: statsmodels (analysis tasks)
+- **Post-compile numpy patch** — ecoscope currently pins `numpy>=2,<2.1`,
+  which conflicts with what conda resolves on some platforms. After compiling,
+  manually edit the generated `pixi.toml` to pin `numpy>=2,<2.1` to match.
+  This constraint will be relaxed in a future ecoscope release.
+- **Post-compile pydantic patch** — ecoscope requires `pydantic<2.9.0` (newer
+  versions break discriminated unions in `apply_classification`), but the
+  compiler emits `pydantic>=2.0.0,<3.0.0` as a conda dependency. After
+  compiling, manually edit the generated `pixi.toml` to pin
+  `pydantic>=2.0.0,<2.9.0`.
```
cisaacstern: Seems like both of these "Post-compile patches" can be included directly in `requirements:` as conda deps with the appropriate pins, similar to `rasterio`?

doc/platform-sdk/content/troubleshooting.md
```diff
@@ -73,6 +73,50 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
...
+- **Separate environment limitation** — `wt-compiler` requires `pydantic>=2.9` while
+  `ecoscope[platform]` requires `pydantic<2.9.0`. They cannot coexist in the
+  same pixi environment. The compiler does not import ecoscope at runtime — it
+  installs it in an ephemeral subprocess environment for task discovery.
```
cisaacstern:
```suggestion
```
This seems irrelevant? wt-compiler uses ephemeral envs for exactly this reason.

doc/platform-sdk/content/troubleshooting.md
```diff
@@ -73,6 +73,50 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
...
+
+- **`python` pin** — without this, the ephemeral discovery environment may
+  resolve Python 3.14+, where transitive deps like `pydantic-core` lack
+  pre-built wheels.
+- **`rasterio` conda dep** — ensures GDAL native libraries are available for
+  the editable ecoscope install.
+- **`extras`** — all three are required for task discovery to succeed:
+    - `platform`: pandera, pydantic, wt-registry, wt-task
+    - `mapping`: lonboard, matplotlib (results/map tasks)
+    - `analysis`: statsmodels (analysis tasks)
```
cisaacstern: I think we should expand this comment to mention that depending on what tasks are used, other extras may also be needed? e.g. `plotting`? Not sure if others may ever be needed or that would be it.

ecoscope/platform/tasks/transformation/_conversion.py
```diff
@@ -36,6 +36,12 @@ def convert_values_to_timezone(
     df: AnyDataFrame,
     timezone: Annotated[str | datetime.tzinfo | TimezoneInfo, Field()],
     columns: Annotated[list[str], Field(description="The columns to convert.")],
+    auto_detect: Annotated[
+        bool,
+        Field(
+            description="Auto-detect all timezone-aware datetime columns to convert, ignoring the columns list.",
+        ),
```
cisaacstern, line 44:
```suggestion
        exclude=True,
        ),
```

Do you ever need this to be exposed in the configuration form / rjsf? Or is it sufficient to only ever set as an inline default in spec.yaml? If inline default is always sufficient, then let's `exclude=True` so that we don't introduce drift in the rjsf of existing workflow configuration forms.

ecoscope/platform/tasks/transformation/_conversion.py
```diff
@@ -44,12 +50,16 @@ def convert_values_to_timezone(
...
+        auto_detect (bool): If True, auto-detect all timezone-aware datetime columns,
+            ignoring the columns list.
 
     Returns:
         AnyDataFrame: The modified DataFrame.
     """
     if isinstance(timezone, TimezoneInfo):
         timezone = timezone.utc_offset
+    if auto_detect:
+        columns = [col for col in df.columns if isinstance(df[col].dtype, pd.DatetimeTZDtype)]
```
cisaacstern, line 65: Preceding this seems like need a check like

```python
if auto_detect and columns:
    raise ValueError("Only one of autodetect and columns may be passed".)
```

... otherwise if someone gives columns + autodetect=True, the autodetect logic will clobber the columns and lead to confusion?

cisaacstern (changes requested):
Some changes requested below inline

doc/platform-sdk/content/troubleshooting.md
Yun-Wu: It's a useful note for myself as a habit to put everything together for testing. It might be worth mentioning here to demonstrate the necessity of the ephemeral environment. Maybe drop the word "limitation"?

doc/platform-sdk/content/troubleshooting.md
Yun-Wu: Good point. I checked that `numpy` can be added in the requirements, but not pydantic as it will be overriden by the compiler.

ecoscope/platform/tasks/transformation/_conversion.py
Yun-Wu: make sense. Thanks!

doc/platform-sdk/content/troubleshooting.md
```diff
@@ -73,6 +73,88 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
...
+    id: workflow_details
+    task: set_workflow_details
+
+  - name: Time Range
+    id: time_range
+    task: set_time_range
+    partial:
+      time_format: '%d %b %Y %H:%M:%S'
+```
+
```
cisaacstern, line 88:
````suggestion
## "Could not read properties of undefined (reading 'name')"

This error appears in Ecoscope Desktop / Ecoscope Web when submitting the configuration form. It means your workflow is missing the required task instance `workflow_details`.

Every ecoscope workflow **must** include:

```yaml
workflow:
  - name: Workflow Details
    id: workflow_details
    task: set_workflow_details
```

````

File: doc/platform-sdk/content/troubleshooting.md [hunk 7]
```diff
@@ -73,6 +73,88 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
...
+**Fix**: Add the missing task(s) to your `spec.yaml`. The `id` values must be exactly `workflow_details` and `time_range` — Ecoscope Desktop and Ecoscope Web look for these specific IDs when processing the configuration form.
```
cisaacstern:
```suggestion
**Fix**: Add the missing task to your `spec.yaml`. The `id` values must be exactly `workflow_details` — Ecoscope Desktop and Ecoscope Web look for this specific ID when processing the configuration form.
```

doc/platform-sdk/content/troubleshooting.md
cisaacstern: Time range is not required anymore, right? I think that was as a change Ray made in response to PGAFF workflows not always needing a time range. Also, error for a malformed time range would be different ("name" is not a field on time range)... my suggestion is we make this section exclusively about workflow_details.

tests/platform/tasks/test_io_smart.py
```diff
@@ -27,6 +27,7 @@ def client():
     ).get_client()
 
 
+@pytest.mark.skip(reason="SMART API is currently down")
```
cisaacstern, line 30: oooh thanks for this! this has been an annoyance on other recent work 🙏

doc/platform-sdk/content/troubleshooting.md
```diff
@@ -73,6 +73,88 @@ The compiler cannot convert a parameter's type to JSON Schema. This happens with
...
+  versions break discriminated unions in `apply_classification`), but the
+  compiler emits `pydantic>=2.0.0,<3.0.0` as a conda dependency. After
+  compiling, manually edit the generated `pixi.toml` to pin
+  `pydantic>=2.0.0,<2.9.0`. This cannot be set via `requirements:` because
+  the compiler overrides pydantic's version constraint.
+- **Separate environments** — `wt-compiler` requires `pydantic>=2.9` while
+  `ecoscope[platform]` requires `pydantic<2.9.0`. They cannot coexist in the
+  same pixi environment, which is why the compiler installs requirements in an
+  ephemeral subprocess environment for task discovery rather than importing
+  them directly.
```
cisaacstern:
```suggestion
```

IMHO this bullet does not add any actionable/useful information to the topic being covered in this section. This section is already quite advanced/intricate, so I would strongly favor not including any extraneous information. Also, this bullet is unnecessarily specific: wt-compiler does not use an ephemeral environment for pydantic incompatibility specifically, but as as hedge against version conflicts more generally (of which pydantic is just one of possibly 100s or 1000s of such cases). I would favor removing this bullet.

cisaacstern (changes requested):
A few notes provided inline regarding accuracy of the docs.

Also (less important) could we perhaps add a note to the title of the PR indicated we are addressing docs as part of this PR as well?

cisaacstern (approved):
LGTM! thanks! and apologies for the delayed review 🙏

PR #684, opened 2026-05-05
feat: polygon event support

### LINKED ISSUE #683: feat: platform task changes for polygon event support
## Problem

`wt-download-events` (and any workflow using these platform tasks) silently centroidizes polygon EarthRanger events, breaks if a non-Point geometry reaches the coord filter, and lacks a built-in way to split a mixed-geometry GDF by geometry type:

1. `ecoscope.platform.tasks.io.get_events` always centroidizes Polygon/MultiPolygon events because `EarthRangerClient.get_events()` defaults `force_point_geometry=True` and the platform task does not surface the flag.
2. `ecoscope.platform.tasks.transformation.apply_reloc_coord_filter`'s inner `envelope_reloc_filter` calls `geometry.x` / `geometry.y` unguarded, which would `AttributeError` on a Polygon.
3. There is no platform task for "filter rows of a GDF by geometry type" — needed at the map-rendering stage to split mixed geometries into separate point/polygon layers.

## Proposal

Three minimal, backwards-compatible platform changes:

1. **`get_events`**: add `force_point_geometry: bool = True` kwarg, forward to `client.get_events(...)`. Default `True` preserves existing centroid behavior.
2. **`apply_reloc_coord_filter` / `envelope_reloc_filter`**: after the existing null-geometry guard, add a simple filter to remove any event polygons that fall completely outside the bounding box filter.
3. **`filter_by_geometry_type`**: new platform transformation. `filter_by_geometry_type(df, geometry_types: list[str])` returns rows whose `geometry.geom_type` is in `geometry_types`. Relocated into platform from the ext-custom task library.

## Refs

- PRD: `wt-download-events/.scratch/ecoscope-prd-wt-download-events-polygons.md` — "Modified Platform Tasks" section + the new-task entry that we decided to host in platform.

Yun-Wu:
## Summary

Three backwards-compatible platform task changes to unblock polygon event support in `wt-download-events`:

1. **`get_events`** (`ecoscope/platform/tasks/io/_earthranger.py`): add `force_point_geometry: bool = True` kwarg, forward to `EarthRangerClient.get_events(...)`. Default `True` keeps existing centroid behavior. Workflows can set `False` to preserve native Polygon/MultiPolygon geometries. Also added the kwarg to `EarthRangerClientProtocol` so mypy is happy.
2. **`apply_reloc_coord_filter`** (`ecoscope/platform/tasks/transformation/_filtering.py`): in `envelope_reloc_filter`, handle non-Point geometries explicitly. Without this, a Polygon would AttributeError on `geometry.x`/`.y`. Polygons/MultiPolygons are kept when they intersect the bounding box (via `geometry.intersects(shapely.geometry.box(...))`) and dropped when disjoint. The exact-coord filter remains Point-only since it doesn't apply to areal geometries.
3. **`filter_by_geometry_type`** (new platform task at `ecoscope/platform/tasks/transformation/_filter_by_geometry_type.py`): filters a GeoDataFrame to rows whose `geometry.geom_type` is in a given list. Relocated from `ecoscope-workflow-task-library` (ext-custom) into the platform package so it ships as a built-in transformation, matching the package boundary of the rest of this work.

## Context

All three changes come from the `wt-download-events` polygon-support PRD ("Modified Platform Tasks" section + the new-task entry, which we decided to host in platform rather than ext-custom). They land together because they're co-dependent at the workflow layer, but each is independently backwards-compatible at the library API.

## Test plan

- [x] `pytest tests/platform/tasks/test_io_earthranger.py::test_get_events_forwards_force_point_geometry` (parametrized True/False) — kwarg is forwarded.
- [x] `pytest tests/platform/tasks/test_io_earthranger.py::test_get_events_force_point_geometry_default_is_true` — locks in the default.
- [x] `pytest tests/platform/tasks/test_io_earthranger.py::test_get_events_empty_response` regression — passes.
- [x] `pytest tests/platform/tasks/test_filtering.py::test_filter_keeps_non_point_geometry_intersecting_bbox` — Polygon overlapping the bbox survives; the exact-coord filter does not drop it.
- [x] `pytest tests/platform/tasks/test_filtering.py::test_filter_drops_non_point_geometry_disjoint_from_bbox` — Polygon disjoint from the bbox is dropped.
- [x] `pytest tests/platform/tasks/test_filtering.py` regression — all 8 tests pass.
- [x] `pytest tests/platform/tasks/test_filter_by_geometry_type.py` — 6 tests covering Point/Polygon/MultiPolygon filtering, empty inputs, CRS/columns preservation.
- [x] `mypy ecoscope/platform/tasks/io/_earthranger.py ecoscope/platform/connections.py ecoscope/platform/tasks/transformation/...` clean.
- [x] Downstream: `wt-download-events` will exercise `force_point_geometry: false` and the new `filter_by_geometry_type` end-to-end via its spec once this is released.

Closes #683

### DIFF
--- a/ecoscope/platform/connections.py
+++ b/ecoscope/platform/connections.py
@@ -115,6 +115,7 @@ class EarthRangerClientProtocol(Protocol):
         include_details: bool,
         include_updates: bool,
         include_related_events: bool,
+        force_point_geometry: bool,
     ) -> AnyGeoDataFrame: ...
 
     def get_event_types(self) -> AnyDataFrame: ...
--- a/ecoscope/platform/tasks/io/_earthranger.py
+++ b/ecoscope/platform/tasks/io/_earthranger.py
@@ -234,6 +234,16 @@ IncludeRelatedEventsAnnotation = Annotated[
         description="Whether or not to include related events",
     ),
 ]
+ForcePointGeometryAnnotation = Annotated[
+    bool,
+    AdvancedField(
+        default=True,
+        description=(
+            "If True, polygon/multipolygon event geometries are reduced to their centroid."
+            " Set False to preserve native geometry types."
+        ),
+    ),
+]
 AnalysisFieldAnnotation = Annotated[
     str,
     Field(
@@ -584,6 +594,7 @@ def get_events(
     include_updates: IncludeUpdatesAnnotation = False,
     include_related_events: IncludeRelatedEventsAnnotation = False,
     include_display_values: IncludeDisplayValuesAnnotation = False,
+    force_point_geometry: ForcePointGeometryAnnotation = True,
 ) -> EventGDF | EventsWithDisplayNamesGDF | EmptyDataFrame:
     """Get events."""
     event_type_ids: list[str] = []
@@ -607,6 +618,7 @@ def get_events(
             include_details=include_details,
             include_updates=include_updates,
             include_related_events=include_related_events,
+            force_point_geometry=force_point_geometry,
         )
     )
 
--- a/ecoscope/platform/tasks/transformation/__init__.py
+++ b/ecoscope/platform/tasks/transformation/__init__.py
@@ -12,6 +12,7 @@ from ._conversion import (
 from ._exploding import explode
 from ._extract import extract_column_as_type, extract_value_from_json_column
 from ._filter import filter_df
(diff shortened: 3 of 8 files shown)


atmorling (approved):
Thanks @Yun-Wu !

PR #686, opened 2026-05-06
fix(earthranger): fall back gracefully when event_type missing from registry

### LINKED ISSUE #685: fix(earthranger): KeyError when event_type is missing from ER registry
## Summary

`EarthRangerIO.get_event_type_display_names_from_events` raises `KeyError` when the events DataFrame contains an `event_type` value that is not present in `get_event_types(include_inactive=True)`. This crashes downstream workflow tasks that depend on the resulting `event_type_display` column.

## Reproduction

Surfaced by an `encounter-rate` workflow run against a westgate ER instance:

```
File ".../ecoscope/io/earthranger.py", line 853, in get_event_type_display_names_from_events
    events_gdf["event_type_display"] = events_gdf["event_type"].map(lambda x: event_type_lookup[x])
File ".../pandas/core/algorithms.py", line 1743, in map_array
    return lib.map_infer(values, mapper, convert=convert)
KeyError: 'test_wildlife_sighting'
```

The event type `'test_wildlife_sighting'` exists on historical events in the database but is no longer registered in ER's event-type catalog (likely deleted/orphaned). The `lambda x: event_type_lookup[x]` formulation does not tolerate that.

## Cause

[`ecoscope/io/earthranger.py:853`](https://github.com/wildlife-dynamics/ecoscope/blob/master/ecoscope/io/earthranger.py#L853) uses a lambda with direct dict indexing:

```python
events_gdf["event_type_display"] = events_gdf["event_type"].map(lambda x: event_type_lookup[x])
```

The neighbouring category lookup at line 875 uses the safer `Series.map(dict)` pattern, which returns NaN for missing keys instead of raising. This bug only affects the display-name lookup.

## Proposed fix

Mirror the `.map(dict)` pattern, then `fillna` with the raw `event_type` so the column stays non-null and downstream groupby/counts behave. The raw type name is preferable to a literal "Unknown" because it preserves the discriminating signal, and preferable to dropping rows because it doesn't mask data.

Yun-Wu:
## Summary
- `EarthRangerIO.get_event_type_display_names_from_events` raised `KeyError` when an event's `event_type` was absent from `get_event_types(include_inactive=True)` (e.g. orphaned/deleted types still attached to historical events).
- Mirror the `Series.map(dict)` pattern already used a few lines below for the category lookup, then `fillna` with the raw `event_type` so the column stays non-null and downstream groupby/counts behave.
- Add a unit test that mocks `get_event_types` to exercise the missing-key path deterministically.
- Update the docstring to note the fallback behavior.

Originally surfaced by a `wg-encounter-rate` live run on the [encounter-rate workflow](https://github.com/wildlife-dynamics/ecoscope-workflows-encounter-rate) (`KeyError: 'test_wildlife_sighting'`), which crashed the downstream pipeline.

Closes #685

### DIFF
--- a/ecoscope/io/earthranger.py
+++ b/ecoscope/io/earthranger.py
@@ -842,6 +842,11 @@ class EarthRangerIO(ERClient):
             If set to duplicates, the event category value will be appended
                 only for event types with overlapping display names
             If set to never, no category names will be appended
+
+        Event types that are present on events but missing from the ER
+        event-type registry (e.g. deleted/orphan types still attached to
+        historical events) fall back to their raw event_type value as the
+        display, rather than raising KeyError.
         Returns
         -------
         gpd.GeoDataFrame
--- a/ecoscope/io/earthranger.py
+++ b/ecoscope/io/earthranger.py
@@ -850,7 +850,9 @@ class EarthRangerIO(ERClient):
 
         event_types = self.get_event_types(include_inactive=True)
         event_type_lookup = dict(zip(event_types["value"], event_types["display"]))
-        events_gdf["event_type_display"] = events_gdf["event_type"].map(lambda x: event_type_lookup[x])
+        events_gdf["event_type_display"] = (
+            events_gdf["event_type"].map(event_type_lookup).fillna(events_gdf["event_type"])
+        )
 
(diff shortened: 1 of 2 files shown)


atmorling (approved):
Thanks @Yun-Wu

PR #691, opened 2026-05-08
Map widget v2

### LINKED ISSUE #618: Allow map definitions to be persisted as deck.gl JSON
We want to be able to serialise a map definition to JSON for the frontend to rehydrate

this will be an `output_type` flag on `draw_map` instructing the task to output a json string rather than html

### LINKED ISSUE #619: Create a `mapV2` widget type
For backwards compatibility in allowing the frontend/server to distinguish between iframe map widgets and the new json definition

This `data` field of this widget will need to ingest the deck gl json definition in order to surface to the result.json of the dashboard

atmorling:
## :earth_americas: Summary
Closes #618
Closes #619

## :package: Proposed Changes
- Adds an `output_type` flag to `draw_map` to output either html or a dict
- Adds a `DeckJsonSpec` BaseModel for some loose type safety around the above output
- Adds a `map_v2` widget, typed off the same `DeckJsonSpec`
- Adds a `persist_json` task for serialising dicts

## 📋 Checklist


ecoscope/platform/tasks/results/_pydeck.py
```diff
@@ -1,18 +1,36 @@
...
+    layers, an initial view state, and a views config) and lets every other key
+    pass through untouched via ``extra="allow"``.
+    """
+
+    model_config = ConfigDict(extra="allow")
+
+    layers: list[dict[str, Any]]
+    initialViewState: dict[str, Any]
+    views: dict[str, Any] | list[dict[str, Any]]
+    widgets: list[dict[str, Any]] = Field(default_factory=list)
```
atmorling, line 31: This is only meant to be a sanity check of `draw_map`'s output contract, and not intended to be a full representation of the deck.gl json.As a future enhancement we could wire up the layer/styling `BaseModel`s defined in this module, but that seemed like overkill for a first pass.

ecoscope/platform/tasks/results/_pydeck.py
```diff
@@ -1039,8 +1062,9 @@ def draw_map(
         map_style=pdk.map_styles.LIGHT_NO_LABELS,
     )
 
-    html = m.to_html(as_string=True)
-    return html
+    if output_type == "json":
+        return DeckJsonSpec.model_validate(json.loads(m.to_json()))
```
atmorling, line 1066: Perhaps this is a bit heavy-handed, but seems easier to work with downstream than returning the raw json string

ecoscope/platform/tasks/results/_widget_types.py
```diff
@@ -5,14 +5,16 @@
 from pydantic_core import Url
 
 from ecoscope.platform.indexes import CompositeFilter
+from ecoscope.platform.tasks.results._pydeck import DeckJsonSpec
```
atmorling, line 8: I feel like this is the right factoring ie, results modules defining their specs and widgets consuming them

atmorling:
Realising now that we don't actually need `persist_json` for https://github.com/wildlife-dynamics/compose/issues/2068. Slightly in favour of leaving it in since this will likely have utility to other SDK users.

cisaacstern (approved):

PR #701, opened 2026-05-14
Harden publish.yml gating to check `wt-registry` cli works as a PR check, and bump pandera + io-core to fix broken `wt-registry` call

### LINKED ISSUE #702: Raising numpy upper bound breaks `wt-registry` cli ability to generate registry output for `ecoscope.platform.tasks`
Following #697, `wt-registry` task schema collection fails on workflow recompiles, with

```
 │ Error: Failed to generate JSON schema for ecoscope.platform.tasks.analysis._aggregation.get_night_day_ratio: Could not convert pandera.typing.pandas.Series[typing.Any] to a NumPy dtype (via `.dtype` value <property object at 0x7efd22a2e930>).

```

This is also not surfaced in any testing here on this repo (it was only caught downstream on recompile attempt).

To fix this I propose we:

- Add testing here to catch this error (to prevent regression on a fix).
- Extensive debugging revealed that this is a subtle pandera version compatibility issue that does not surface easily in the pip-built unit test envs, so we should instead implement this test directly on the rattler build recipe. This will be the most faithful reproducer.
- Fix the pandera pin to resolve this issue

cisaacstern:
## :earth_americas: Summary

closes #702

## :package: Proposed Changes

- Split build-for-prefix into build-only and publish-to-prefix jobs; build now runs on pull_request so recipe breakage is caught at PR time.
- Make github-release depend on publish-to-prefix so a failed prefix upload blocks the GH release, keeping GH and prefix.dev in sync.
- Add a wt-registry --format json script test to the ecoscope-platform recipe output to guard against entry-point regressions.

## 🔬 Verification

- Note that the compiler error reported in #702 was initially reproduced in https://github.com/wildlife-dynamics/ecoscope/actions/runs/25838427138/job/75918494073#step:8:2177
- The fix for this was raising the pandera lower bound to 0.28 for numpy compat, and this in turn required raising the io-core lower bound (to accomodate the higher pandera)
- With those changes made, the build test check called out in https://github.com/wildlife-dynamics/ecoscope/pull/701#discussion_r3238998908 is now passing https://github.com/wildlife-dynamics/ecoscope/actions/runs/25840708716/job/75925226153

## 📋 Checklist

### DIFF
--- a/publish/recipes/release/ecoscope.yaml
+++ b/publish/recipes/release/ecoscope.yaml
@@ -79,7 +79,7 @@ outputs:
         - wt-registry >=0.2.0,<0.3.0
         - wt-task >=0.1.0,<0.2.0
         - pydeck ==0.9.2
-        - ecoscope-earthranger-io-core ==0.0.12
+        - ecoscope-earthranger-io-core ==0.0.14
     tests:
       - python:
           imports:
--- a/publish/recipes/release/ecoscope.yaml
+++ b/publish/recipes/release/ecoscope.yaml
@@ -72,7 +72,7 @@ outputs:
     requirements:
       run:
         - ecoscope ==${{ version }}
-        - pandera >=0.24.0,<0.29
+        - pandera >=0.28,<0.29
         - pydantic <2.9.0
         - pydantic-settings
         - cloudpathlib-gs
--- a/publish/recipes/release/ecoscope.yaml
+++ b/publish/recipes/release/ecoscope.yaml
@@ -79,7 +79,7 @@ outputs:
         - wt-registry >=0.2.0,<0.3.0
         - wt-task >=0.1.0,<0.2.0
         - pydeck ==0.9.2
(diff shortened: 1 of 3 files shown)


publish/recipes/release/ecoscope.yaml
```diff
@@ -72,18 +72,21 @@ outputs:
...
         - pydeck ==0.9.2
-        - ecoscope-earthranger-io-core ==0.0.11
+        - ecoscope-earthranger-io-core ==0.0.14
     tests:
       - python:
           imports:
             - ecoscope.platform
+      - script: |
+          wt-registry --package ecoscope.platform.tasks --format json \
+            | python -c "import json, sys; data = json.load(sys.stdin); assert data, 'wt-registry returned empty result'"
```
cisaacstern, line 89: Here's the test that reproduces the compiler error, run as part of rattler build, because the specific dep combination that triggered it was very hard to reproduce reliably on the pip side.

PR #706, opened 2026-05-18
Improve test execution runtime

### LINKED ISSUE #662: Reduce execution time of non-io, non-platform unit tests
The core library (non-platform) tests take ~8m to complete which is way too long. Seems like the main culprit [here](https://github.com/wildlife-dynamics/ecoscope/actions/runs/24046748238/job/70131940501) is the ecograph->geotiff tests

atmorling:
## :earth_americas: Summary
Closes #662

## :package: Proposed Changes
- Refactors / removes a bunch of problem test to speed up overall runtime

## 📋 Checklist
- [ ] ~~Corresponding ecoscope.platform tasks updated if necessary~~


ecoscope/analysis/UD/etd_range.py
```diff
@@ -113,6 +113,8 @@ def calculate_etd_range(
     """
 
     trajectory_gdf = trajectory.gdf if isinstance(trajectory, Trajectory) else trajectory
+    if weibull_pdf is None:
+        weibull_pdf = Weibull2Parameter()
```
atmorling, line 117: With the reworked tests we actually exposed a mutable default here

tests/test_earthranger_io.py
atmorling, line 1: Removing a bunch of redundant parity tests.. kwargs parity is already covered by other tests so these were just hitting the ER API unnecessarily

tests/test_ud.py
atmorling, line 1: Reworked tests here so that rather than testing etd against the full movebank dataset twice, we use a smaller mock trajectory to test the plumbing of `calculate_etd_range` (ie; when `output_path=None` check that no file is written)

cisaacstern (approved):

PR #713, opened 2026-05-22
Low-hanging fruit test coverage

### LINKED ISSUE #661: Improve test coverage in the specific modules where it's lacking
Our overall test coverage is around 85%, however within specific modules, like Seasonal analysis and EE tools, for example, the coverage is much lower, between 20-50%.

atmorling:
## :earth_americas: Summary
Closes #661 

## :package: Proposed Changes
- Adds some low-hanging fruit test coverage to modules that were lacking.
- This hasn't made as much of an overall dent as expected since our coverage has steadily lifted since the time #661 was written

## 📋 Checklist


ecoscope/io/utils.py
```diff
@@ -38,6 +38,10 @@ def download_file(
     the response header
     """
 
+    if not os.path.isdir(path) and os.path.exists(path) and not overwrite_existing:
+        print(f"{path} exists and overwrite_existing is False. Skipping...")
+        return
```
atmorling, line 43: Previously we'd still download the file even if we weren't going to overwrite the path

PR #786, opened 2026-07-24
fix: migrate Convert Units conditional from legacy dependencies to allOf/if/then

### LINKED ISSUE #785: fix: Convert Units conditional uses legacy JSON Schema dependencies keyword
StatSummaryParam (compiled into workflow schemas as the Custom summary metric) reveals its unit-conversion fields via the legacy \`dependencies\` keyword in \`json_schema_extra\`. \`dependencies\` was split into \`dependentSchemas\`/\`dependentRequired\` in JSON Schema 2019-09 and removed from the vocabulary — ecoscope-server validates submitted formdata with \`Draft202012Validator\`, which silently ignores the keyword, so the conditional (units only meaningful when Convert Units is checked, values restricted to the Unit enum) is unenforced server-side.

Same dialect problem as wildlife-dynamics/wt-download-subject-tracks#29, where it additionally caused 422 rejections. Here there is no \`additionalProperties: false\` wall, so nothing rejects — but the constraint is dead weight to the validator.

Fix: migrate to \`allOf\`/\`if\`/\`then\`, which both the RJSF renderer (conditional reveal, identical UX) and the 2020-12 validator understand.

Yun-Wu:
## Summary
- Rewrite `StatSummaryParam`'s `json_schema_extra` conditional (Convert Units → Original/New Unit reveal) from the legacy `dependencies`/`oneOf` keyword to `allOf`/`if`/`then`
- Rename + update the schema-shape test accordingly (also asserts `new_unit` options now)

## Why
`dependencies` was removed from the JSON Schema vocabulary in 2019-09. ecoscope-server validates submitted formdata against the compiled rjsf with `Draft202012Validator`, which silently ignores the keyword — so the units conditional was unenforced server-side. `allOf`/`if`/`then` is understood by **both** the RJSF renderer and the 2020-12 validator. Same dialect fix as wildlife-dynamics/wt-download-subject-tracks#31, which has the full analysis; unlike that case there's no `additionalProperties: false` on this model, so nothing was being rejected — this closes the enforcement gap and retires the dead keyword.

## Verification
- `tests/platform/tasks/test_patrol_summary.py` — 11 passed
- Renderer parity against ecoscope-web's installed `@rjsf` packages (`retrieveSchema` + the app's exact ajv config): unit fields hidden when unchecked, both labeled unit selects revealed when checked, zero ajv errors for valid and orphaned-unit form states
- `Draft202012Validator`: now enforces the Unit enum when Convert Units is checked (previously ignored), tolerates orphaned unit values left by unchecking (matching the runtime `check_units` validator, which nulls them)

Closes #785

### DIFF
--- a/ecoscope/platform/tasks/analysis/_summary.py
+++ b/ecoscope/platform/tasks/analysis/_summary.py
@@ -28,40 +28,41 @@ class _BaseSummaryParam(BaseModel):
 
 
 # The unit fields are excluded from the model's own JSON schema (SkipJsonSchema)
-# and re-introduced via the `dependencies` block in json_schema_extra, so the form
-# only shows them once "Convert Units" is checked. Note the docstring renders as
-# the form's helper text.
+# and re-introduced via the allOf/if/then block in json_schema_extra, so the form
+# only shows them once "Convert Units" is checked. if/then (not the legacy
+# `dependencies` keyword) is understood by BOTH the RJSF renderer and JSON Schema
+# 2020-12 validators (ecoscope-server validates submitted formdata with
+# Draft202012Validator, which silently ignores `dependencies`). See
+# wildlife-dynamics/wt-download-subject-tracks#31 for the full dialect story.
+# Note the docstring renders as the form's helper text.
 class StatSummaryParam(_BaseSummaryParam):
     """Pick a column and statistic, with optional unit conversion."""
 
     model_config = ConfigDict(
         title="Statistic",
         json_schema_extra={
-            "dependencies": {
-                "convert_units": {
-                    "oneOf": [
-                        {"properties": {"convert_units": {"const": False}}},
-                        {
-                            "properties": {
-                                "convert_units": {"const": True},
-                                # No `default` on these: a default on a dependency branch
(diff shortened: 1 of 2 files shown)


atmorling (approved):
Thanks @Yun-Wu, have approved as the diff looks good but I'm not in love with how this is documented... there's really no collateral/anything to back up the claims from the Verification section of the PR body:

>- Renderer parity against ecoscope-web's installed @rjsf packages (retrieveSchema + the app's exact ajv config): unit fields hidden when unchecked, both labeled unit selects revealed when checked, zero ajv errors for valid and orphaned-unit form states 

It would be good to see either a link to a test branch where these changes have been compiled against (so we can compare the generated RJSF), _or_ just to remove this note entirely


>- Draft202012Validator: now enforces the Unit enum when Convert Units is checked (previously ignored), tolerates orphaned unit values left by unchecking (matching the runtime check_units validator, which nulls them)

I think this is referring to changes in `Draft202012Validator` but makes it sound like _this_ PR made the changes which is obviously not the case. I think we should remove this.

PR #788, opened 2026-07-24
feat: add "Normalised (LTD)" patrol density weighting with mode-aware classification

### LINKED ISSUE #787: feat: add "Normalised (LTD)" patrol density weighting option
The patrol-track-density-map workflow (split from patrol-efforts) has a single "Density Calculation" dropdown driven by `set_patrol_weighting_spec`, currently offering Time / Distance. Add a third option, "Normalised (LTD)", that reproduces the patrols workflow's Time Density Map: `calculate_ltd` (per-cell timespan sum / total time) followed by percentile classification with `LTD_DEFAULT_PERCENTILES` and dissolve by percentile.

Because the workflow DAG is static, one dropdown must drive all three modes through one spec pipeline. Time/Distance use equal-interval value bins while LTD uses percentile bins, so classification moves inside a mode-aware composite task that emits a uniform output shape (`density` numeric display value + `density_bins` string labels).

Scope:
- `Unit.PERCENT` display-only unit
- `WeightingSpec.mode` ("sum"/"ltd") + composite task `calculate_classified_track_density`
- Third `PATROL_WEIGHTING_SPECS` entry keyed independently of the sum column
- patrol-efforts keeps working unchanged; patrol-track-density-map is rewired to the composite task separately

Yun-Wu:
## Summary
- Add a third patrol density weighting option, **Normalised (LTD)**, to `set_patrol_weighting_spec`, reproducing the patrols workflow's Time Density Map (`calculate_ltd` → percentile classification → dissolve by percentile). Percentile levels are user-configurable: the picker is revealed only when Normalised (LTD) is selected (RJSF `dependencies` pattern) and comes pre-filled with the LTD defaults (patrols parity); orphaned percentiles left behind by switching modes are discarded server-side.
- New mode-aware composite task `calculate_classified_track_density(geodataframe, meshgrid, weighting_spec)` emits a uniform shape across modes — numeric `density` (hours/km for sum modes, percentile for UD) + ready-to-render `density_bins` labels (UD labels bake in the `" %"` suffix) — so one static spec pipeline can drive all three dropdown modes. The sum path is numerically identical to the existing `calculate_feature_density → normalize_density_units → sort_values → drop_nan_values_by_column → apply_classification` chain (tested).
- Split `WeightingSpec` into two frozen dataclasses under a union alias: `SumWeightingSpec` (sum column, original/display units, equal-interval bins) and `UDWeightingSpec` (percentiles, `%` display unit) — LTD is just the current UD algorithm, leaving room for others. Each carries a fixed `Literal` mode tag (`"sum"`/`"ud"`) as the union discriminator; dispatch is `isinstance`-based. `normalize_density_units` and `get_weighting_column` narrowed to `SumWeightingSpec`. The composite task and its branch helpers moved to a new `_track_density.py`, leaving `_density_weighting.py` as the spec layer (dataclasses + small spec-reading tasks).
- New `get_density_colormap` task: sum weightings use `RdYlGn_r` (high sums red); UD flips to `RdYlGn` since the lowest percentile isopleth is the densest core. The UD legend title is "Time Spent (%)"; `Unit.PERCENT` added as a display-only unit.
- `calculate_feature_density`'s sum mode made optional and robust; add `set_optional_string_var` config task.

Closes #787

### DIFF
--- a/ecoscope/platform/tasks/analysis/_density_weighting.py
+++ b/ecoscope/platform/tasks/analysis/_density_weighting.py
@@ -29,9 +29,7 @@ class UDWeightingSpec:
     legend_label: str | None = None  # legend title prefix when it differs from option_label
     percentiles: tuple[float, ...] | None = None  # percentile bins; None -> LTD defaults
     display_unit: Unit = Unit.PERCENT  # unit shown on the map and in the legend title
-    # Percentiles are emitted ascending and the lowest isopleth is the densest
-    # core, so it gets the first color: red -> green (matches patrols' Time
-    # Density Map).
+    # Lowest isopleth is the densest core, so red comes first (patrols parity).
     colormap: str = "RdYlGn"
     mode: Literal["ud"] = "ud"  # union discriminator
 
--- a/ecoscope/platform/tasks/analysis/_density_weighting.py
+++ b/ecoscope/platform/tasks/analysis/_density_weighting.py
@@ -1,11 +1,11 @@
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Annotated, Literal
+from typing import Annotated, Literal, cast
 
 from pydantic import Field
 from wt_registry import register
 
-from ecoscope.platform.annotations import AnyGeoDataFrame
+from ecoscope.platform.annotations import AnyDataFrame, AnyGeoDataFrame
 from ecoscope.platform.tasks.transformation._unit import Unit, with_unit
 
 
@@ -102,9 +102,11 @@ def _classified_sum_density(
         apply_classification,
     )
 
+    # copy: calculate_feature_density writes into the grid it is given, and the
+    # same meshgrid object is shared across mapvalues groups
     result = calculate_feature_density(
         geodataframe=geodataframe,
-        meshgrid=meshgrid,
+        meshgrid=meshgrid.copy(),
         geometry_type="line",
         sum_column=weighting_spec.density_sum_column,
     )
@@ -114,12 +116,15 @@ def _classified_sum_density(
     if result.empty:
         return _empty_density_result(meshgrid)
 
-    return apply_classification(
-        df=result,
-        input_column_name="density",
-        output_column_name="density_bins",
-        label_options=DefaultLabels(label_ranges=True, label_decimals=1),
-        classification_options=SharedArgs(scheme="equal_interval", k=10),
+    return cast(
+        AnyGeoDataFrame,
+        apply_classification(
+            df=cast(AnyDataFrame, result),
+            input_column_name="density",
+            output_column_name="density_bins",
+            label_options=DefaultLabels(label_ranges=True, label_decimals=1),
+            classification_options=SharedArgs(scheme="equal_interval", k=10),
+        ),
     )
 
 
@@ -138,10 +143,10 @@ def _classified_ltd_density(
 
     result = classify_percentile(
         df=density_grid,
-        percentile_levels=[float(p) for p in LTD_DEFAULT_PERCENTILES],
+        percentile_levels=[float(p) for p in LTD_DEFAULT_PERCENTILES],  # type: ignore[arg-type,misc]
         input_column_name="density",
     )
-    result = result.dissolve(
+    result = result.dissolve(  # type: ignore[operator]
         "percentile",
         aggfunc={"density": "sum"},
         as_index=False,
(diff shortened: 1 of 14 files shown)


ecoscope/platform/tasks/config/_set_vars.py
```diff
@@ -14,6 +14,13 @@ def set_string_var(
     return var
 
 
+@register()
+def set_optional_string_var(
+    var: Annotated[str, Field(title="")] = "",
+) -> str:
+    return var
```
atmorling, line 21: Dumb question but how does this work at the spec level? Is this exposed as user configurable?

ecoscope/platform/tasks/analysis/_density_weighting.py
```diff
@@ -10,11 +10,31 @@
...
+    option_label: str  # form dropdown label and default legend title prefix
+    legend_label: str | None = None  # legend title prefix when it differs from option_label
+    percentiles: tuple[float, ...] | None = None  # percentile bins; None -> LTD defaults
+    display_unit: Unit = Unit.PERCENT  # unit shown on the map and in the legend title
+    # Lowest isopleth is the densest core, so red comes first (patrols parity).
+    colormap: str = "RdYlGn"
+    mode: Literal["ud"] = "ud"  # union discriminator
+
+
+WeightingSpec: TypeAlias = SumWeightingSpec | UDWeightingSpec
```
atmorling: nit: the `mode` discriminator on `SumWeightingSpec` and `UDWeightingSpec` is unnecessary. If I understand this correctly (and please stop me if I've missed something), the union is resolvable from the difference in field names alone

ecoscope/platform/tasks/analysis/_patrol_density.py
```diff
@@ -1,35 +1,116 @@
...
+    model_config = ConfigDict(
+        title="",
+        json_schema_extra={
+            "dependencies": {
+                "density_sum_column": {
+                    "oneOf": [
+                        {
+                            "properties": {
+                                "density_sum_column": {"enum": ["timespan_seconds", "dist_meters"]},
+                            }
```
atmorling: This needs to be updated from `dependencies` to the  `allOf` / `if/then` form

ecoscope/platform/tasks/transformation/_unit.py
```diff
@@ -17,6 +17,8 @@ class Unit(Enum):
     DAY = "d"
     METERS_PER_SECOND = "m/s"
     KILOMETERS_PER_HOUR = "km/h"
+    # Display-only label; never routed through astropy conversion.
+    PERCENT = "%"
```
atmorling: IIUC there's nothing that _enforces_ that `PERCENT` is never routed through a unit conversion, and the conversion would raise if one tried to convert, say, `PERCENT` to `DAY`. The behaviour makes sense IMO, but this comment is a bit misleading.

atmorling (changes requested):
Thanks @Yun-Wu, left a couple minor comments but have requested changes on account of the `dependencies` json schema block in `PatrolWeightingSelection`

ecoscope/platform/tasks/config/_set_vars.py
Yun-Wu: yes it is used for an optional aggregate column when calculating sum in a heatmap

ecoscope/platform/tasks/analysis/_density_weighting.py
Yun-Wu: You're right — nothing dispatches on `mode` (dispatch is `isinstance`-based), and I verified pydantic resolves the union from field shape alone on a JSON round-trip, in either union order (a UD payload lacks Sum's required fields; the Sum payload's extra keys keep UD from claiming it). Removed in ec8bdf2.

ecoscope/platform/tasks/analysis/_patrol_density.py
Yun-Wu: Sorry for the miss here. Converted to `allOf`/`if`/`then` in ec8bdf2 (the sum-modes branch contributed nothing, so a single conditional suffices).

ecoscope/platform/tasks/transformation/_unit.py
Yun-Wu: Fair — reworded in ec8bdf2 to say it's display-only by convention and that astropy would raise for non-dimensionless targets, rather than claiming it's never converted.

atmorling (approved):
TYSM @Yun-Wu

PR #790, opened 2026-07-29
Create Parent Dirs for Nested Filenames in `_get_path`

### LINKED ISSUE #791: Attachments With Nested Filenames Fail to Download
Attachments uploaded from mobile phone include the full upload path in the filename (for example, `uuid/image_fileuploads/2026/7/28/uuid/photo.jpg`) instead of just the basename. The `_get_path` function only creates the root directory. This results in a `FileNotFoundError`, causing the entire `download_event_attachments` task to fail.

zakhusayn:
## :earth_americas: Summary
Attachments uploaded to Events from mobile phones return a filename that includes the full upload path (e.g. `uuid/image_fileuploads/2026/7/28/uuid/photo.jpg`) rather than a plain basename. `_get_path` only created local_path (the root dir), so when write_bytes was called on a nested path, it raised **FileNotFoundError**, causing `download_event_attachments` to fail.

## :package: Proposed Changes
Construct write_path first, then call `write_path.parent.mkdir(parents=True, exist_ok=True)` it handles both plain filenames and nested paths

Closes #791

### DIFF
--- a/ecoscope/platform/serde.py
+++ b/ecoscope/platform/serde.py
@@ -46,9 +46,8 @@ def _get_path(root_path: str, filename: str):
             else:
                 # Standard file URL or local path
                 local_path = Path(parsed_url.path)
-            if not local_path.exists():
-                local_path.mkdir(parents=True, exist_ok=True)
             write_path = local_path / filename
+            write_path.parent.mkdir(parents=True, exist_ok=True)
             read_path = write_path.absolute().as_posix()
 
         case "gs":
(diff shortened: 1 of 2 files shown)


atmorling:
Just noting for posterity that I pushed back on the fix as implemented in this PR via a  Slack sync with @zakhusayn.
- We can fix this bug from the task by parsing the filename returned from the ER API
- The serde change in this PR makes the average case file persistence more permissive by allowing arbitrary folder structures in the results bucket, which at least of the time of writing, we do not want to be doing.

PR #749, opened 2026-06-24
Patrol summary metrics & density-weighting tasks

Yun-Wu:
## Summary
Builds out the tasks behind the patrol-effort dashboard: a configurable patrol summary table, time/distance-weighted density heatmaps, a ground-coverage area metric, and per-layer legend titles on ecomaps. Also refactors `summarize_df` params into a discriminated union so each metric type exposes only the fields that apply to it.

### Patrol summary
- `set_patrol_summary_metrics`: preset-driven summary table. Ships Patrol Count, Total Distance, Total Duration, Patrol Days, and Merged/Unmerged Area Covered presets (each knows its column/statistic, so the form only asks what varies — target unit, swath width), plus a **Custom** escape hatch exposing the full statistic form.
- `summarize_df`: `SummaryParam` is now a discriminated union on `aggregator`:
  - `StatSummaryParam` — column statistic (sum/count/min/max/mean/nunique/median) with optional unit conversion hidden behind a "Convert Units" checkbox.
  - `UniqueSummaryParam` — a column's distinct values as a comma-separated list.
  - `CoverageSummaryParam` — ground area covered (no column).
  - `NightDayRatioSummaryParam` — night/day fix ratio (no column/units).
  - Legacy specs passing `original_unit`/`new_unit` without `convert_units` still convert (back-compat).

### Coverage area
- `coverage_area_km2` (`ecoscope.base.utils`): buffers track segments by `swath_width_meters / 2` per side and returns covered ground area in km². Takes an optional `crs`; by default it computes in the data's local UTM zone. `merged=True` → union area (overlaps counted once); `merged=False` → sum of per-segment corridor areas. Exposed via the `coverage_area` aggregator on `summarize_df`.

### Density weighting
- `WeightingSpec` (`_density_weighting.py`, generic): bundles a weighting's source column, original/display units, and label. The generic tasks `normalize_density_units` (raw density → display units), `get_density_legend_title` (label + display unit, e.g. "Time (h)"), and `get_weighting_column` (feeds `calculate_feature_density.sum_column`) all consume a `WeightingSpec`, so they carry no patrol assumptions.
- `set_patrol_weighting_spec` (`_patrol_density.py`, patrol-specific): the one patrol-aware task — a labeled Time/Distance dropdown over `timespan_seconds`/`dist_meters` that returns the matching `WeightingSpec`.
[...]

PR #798, opened 2026-07-30
allow `decimal_places=None` in StatSummaryParam for Timestamp aggregations

### LINKED ISSUE #797: `StatSummaryParam.decimal_places` should accept None for non-numeric aggregations on Timestamp columns
`StatSummaryParam.decimal_places` is typed as strictly `int`, which means that setting `decimal_places: null` in a spec for `min`/`max` aggregations on `Timestamp` columns (e.g. patrol start/end times) fails Pydantic validation at startup. The workaround of setting it to `0` instead causes a runtime crash because the code calls `round(timestamp, 0)`, and `Timestamp` objects do not support `__round__`. The runtime already handles `None` correctly by skipping rounding when `decimal_places` is `None`, so the only issue is that the type annotation is too strict.

zakhusayn:
Closes #797

### DIFF
--- a/ecoscope/platform/tasks/analysis/_summary.py
+++ b/ecoscope/platform/tasks/analysis/_summary.py
@@ -59,7 +59,7 @@ class StatSummaryParam(_BaseSummaryParam):
     )
     aggregator: Annotated[AggOperations, Field(title="Statistic")]
     column: Annotated[str, Field(title="Column", description="Column to aggregate.")]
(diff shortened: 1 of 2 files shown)


atmorling (commented):
LGTM thanks @zakhusayn - This _looks_ like a miss from https://github.com/wildlife-dynamics/ecoscope/pull/749 I'm going to defer approval to @Yun-Wu to be sure

Yun-Wu (approved):
TYSM Zak!

PR #793, opened 2026-07-29
feat: combined patrol download config task (set_patrol_download_params)

Yun-Wu:
## Summary
- Add `set_patrol_download_params` to `ecoscope.platform.tasks.config`: combined patrol track + event download configuration (filetypes for both branches, AdvancedField filename prefixes defaulting to `patrol_tracks` / `patrol_events`), returning a 4-tuple.
- Add scalar splitters `get_patrol_track_filetypes` / `get_patrol_track_filename_prefix` / `get_patrol_event_filetypes` / `get_patrol_event_filename_prefix` — the workflow spec template engine cannot subscript returns.
- Mirrors the existing `set_download_params` (wt-download-subjects) pattern; consumed by the wt-download-patrols merged 'Patrol Track & Event Download File Format' card.
- Tests: `tests/platform/tasks/test_set_patrol_download_params.py` (6 tests).

## Unrelated CI deflake
- Seed the GAMM MCMC sampler in tests: `test_gamm_regressor_population_predict` intermittently failed because the shared `fitted_gamm` fixture ran unseeded MCMC (1 chain, 100 draws/tune), occasionally estimating the site random intercepts poorly. `GAMMRegressor` now accepts a `random_seed` passed through to the sampler, and the fixture sets `random_seed=42`.

Closes #792

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #801, opened 2026-07-30
ERDW-247: Request subject `additional` from the DWH when needed

### LINKED ISSUE #800: Subject colors are lost when observations come from the DWH
`assign_subject_colors` reads each subject's `rgb` from `subject__additional`. The EarthRanger API serves that field with subject details, but the warehouse leaves it null unless a query explicitly asks for it — so the same workflow yields per-subject track colors on the ER API and a single fallback color on the DWH.

The warehouse now exposes an opt-in `include_subject_additional` param and we need to set it when needed in ecoscope tasks.

marianobrc:
## :earth_americas: Summary
Closes #800 
## :package: Proposed Changes
- Adds an opt-in `include_subject_additional` param to the `get_subjectgroup_observations` task, forwarded only on the warehouse path (the  EarthRanger API already serves the field unconditionally), and bumps the io-core pin to `v0.0.22 `where the flag exists.
- Also declares `extra__subject__additional` on the subject-group schema as optional and nullable.

## 📋 Checklist

### DIFF
--- a/ecoscope/platform/tasks/io/_earthranger.py
+++ b/ecoscope/platform/tasks/io/_earthranger.py
@@ -481,8 +481,23 @@ def get_subjectgroup_observations(
             description="Filter observations based on exclusion flags.",
         ),
     ] = "clean",
+    include_subject_additional: Annotated[
+        bool,
+        AdvancedField(
+            default=False,
+            title="Include Subject Additional",
+            description="Whether or not to include the subject's free-form `additional` JSON",
+        ),
+    ] = False,
 ) -> SubjectGroupObservationsGDF | EmptyDataFrame:
-    """Get observations for a subject group from EarthRanger."""
+    """Get observations for a subject group from EarthRanger.
+
+    ``include_subject_additional`` only affects the warehouse path. The EarthRanger API
+    path serves the subject's ``additional`` JSON unconditionally (it rides along with
+    ``include_subject_details``), whereas the warehouse serves the column null unless
+    asked, so consumers of ``subject__additional`` -- e.g. ``assign_subject_colors``,
+    which reads its ``rgb`` key -- must opt in to get the same data from both backends.
+    """
     from ecoscope.relocations import Relocations
 
     filter_int = _EXCLUSION_FILTER_TO_INT[filter]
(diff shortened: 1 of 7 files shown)


ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -481,8 +481,23 @@ def get_subjectgroup_observations(
...
         ),
     ] = "clean",
+    include_subject_additional: Annotated[
+        bool,
+        AdvancedField(
+            default=False,
+            title="Include Subject Additional",
+            description="Whether or not to include the subject's free-form `additional` JSON",
+        ),
+    ] = False,
```
atmorling, line 491:
```suggestion
```
Since the internals of whether ER IO tasks hit the warehouse or not are meant to be relatively opaque to workflow authors, I don't think this needs to be a configurable param on the task and can probably just be fixed to always be `True`

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -502,6 +517,7 @@ def get_subjectgroup_observations(
             since=time_range.since.isoformat(),
             until=time_range.until.isoformat(),
             filter=filter_int,
+            include_subject_additional=include_subject_additional,
```
atmorling, line 520:
```suggestion
            include_subject_additional=True,
```

atmorling (changes requested):
Thanks @marianobrc - just the one note about whether `include_subject_additional` needs to be exposed as a flag at the task level (as opposed to being `True` in all cases for `get_subjectgroup_observations`).
I've requested changes because I feel like it's not strictly necessary unless there's a good reason that I'm missing.

Also noting that the GAMM test failing in CI is not related to this PR and will be fixed once #793 is merged.

marianobrc:
> Thanks @atmorling  - just the one note about whether `include_subject_additional` needs to be exposed as a flag at the task level (as opposed to being `True` in all cases for `get_subjectgroup_observations`). I've requested changes because I feel like it's not strictly necessary unless there's a good reason that I'm missing.
> 
> Also noting that the GAMM test failing in CI is not related to this PR and will be fixed once #793 is merged.

Thanks @atmorling, good question. The reason why I wanted to make this an opt-in feature is that retrieving subject `additional` json field has a cost in terms of performance (larger payload -> longer data transfer time)  and memory usage, and as far as I know, only one workflow using this task needs additional data (download subject tracks), hence the default is false. When using the ER API you are already paying that cost and there's no way to exclude `additional` but we are on time to support this when using the warehouse, and I think it also aligns better with the main motivation for using the DWH: better performance.

atmorling (approved):
Thanks for the clarification @marianobrc  - agreed if there's perf implications let's leave the flag in

PR #803, opened 2026-08-01
Add transformation tasks and rework encounter-rate metrics

### LINKED ISSUE #802: feat: add concat_dataframes transformation task
Add a `concat_dataframes` task to `ecoscope.platform.tasks.transformation` for row-wise concatenation of 2+ dataframes that:

- Drops skipped upstream dependencies (`SkippedDependencyFallback` / skip sentinels, same pattern as `groupbykey`) so a skipped branch degrades gracefully
- Ignores empty inputs, returning an empty `DataFrame` when all inputs are dropped/empty so downstream `any_is_empty_df` skips cleanly
- Preserves `GeoDataFrame` type and CRS when any input is a GeoDataFrame
- Optionally guarantees columns exist in the output via `ensure_columns` (added with null values if absent) so downstream aggregations never hit a missing column
- Supports `reset_index` (default true)

Needed by the `patrol-encounter-rate-stats` workflow (combine patrol/event branches before split & summary).

Yun-Wu:
## Summary

Transformation tasks and an encounter-rate metrics rework — the pieces consumed together by the `patrol-encounter-rate-stats` workflow (combining patrol/event branches, then computing rate metrics). The trend chart built on top of these lands separately in a stacked PR.

### Transformation tasks (`ecoscope.platform.tasks.transformation`)

- **`concat_dataframes`** — row-wise concatenation of 2+ dataframes, modeled after `groupbykey`'s skip-sentinel handling (`SkippedDependencyFallback` drops skipped upstream dependencies before validation). Empty inputs are ignored; if all inputs are empty, the result is an empty frame preserving the union of input columns (a bare empty `DataFrame` only when every input was skipped), so downstream `any_is_empty_df` skips cleanly without losing the schema. Preserves `GeoDataFrame` type and CRS when any input is a GeoDataFrame (including mixed df/gdf inputs where `pd.concat` would otherwise return a plain frame). `ensure_columns` adds null-valued columns absent after concatenation so downstream aggregations never hit a missing column.
- **`align_keyed_iterable_to_reference`** (moved from ecoscope-workflow-task-library, see wildlife-dynamics/ecoscope-workflow-task-library#225) — reindexes a keyed iterable of DataFrames to the key universe of a reference keyed iterable, filling missing keys with an empty schema-matching DataFrame (or an explicit `fill_value`); target-only keys are preserved. Duplicate target keys and an empty target without `fill_value` raise `ValueError` instead of silently dropping data or producing schema-less frames.
- **`decompose_datetime`** — ported from `ecoscope-workflows-ext-custom` so platform-only workflows can derive calendar columns (notably `segment_start_date` for the per-patrol-day rate metric); `week`/`weekofyear` components map to `isocalendar().week` (removed from the `.dt` accessor in pandas 2.0). ⚠️ Registered under a different FQN than the ext-custom copy — specs installing both must reference it by dotted path.
- **`extract_grouper_index_names` / `rename_grouper_index_columns`** — grouper index helpers; `add_temporal_index` recomputes an existing grouper index level from the current `time_col` (idempotent, never stale); `add_spatial_index` skips groupers whose index level already exists.

### Encounter-rate metrics rework (`ecoscope.platform.tasks.analysis`)

- **`RatioOperand` / `RatioSummaryParam`** — per-group ratio of two aggregations (ratio-of-sums, not mean-of-ratios) with a `scale` factor; NaN when the denominator is 0.
- **Count/Sum aggregation annotation** (`CountAggregation`, `SumOfColumnAggregation`, `make_aggregation_json_schema_extra`, colocated with the encounter-rate task in `_patrol_summary.py`) — emits a flat `allOf`/`if`/`then` form schema instead of an `anyOf`, so RJSF retains a typed column when the user toggles between Count and Sum modes.
- **`set_encounter_rate_metrics`** — produces `list[SummaryParam]` from a task-level aggregation ("Measure Encounters By": Number of Events | Sum of an Event Field) plus a list of `EncounterRateMetric` rows:
  - New metrics: Total Events (always counts), Total Encounters, Encounters per Duration (h|d), per Distance (km|m), per Patrol, per Patrol Day — alongside the full patrol preset union (count, days, distance, duration, area covered, custom). Union order is alphabetical by title (drives the dropdown).
  - Encounter metrics follow the task-level count/sum mode; each row's optional "Event Field to Sum" overrides the task-level field (mirroring the encounter-rate map task's wording and semantics).
  - Column labels are derived from the aggregate column ("Number of Animals per Hour"); with no column they collapse to "Events per Hour", "Events per Km", "Total Events", etc.
  - Per-patrol / per-patrol-day rates use `nunique` denominators (`patrol_id` / `segment_start_date` — the latter is why `decompose_datetime` was ported).

### Tests

100 tests: `test_concat`, `test_align_keyed_iterable`, `test_decompose_datetime`, `test_indexing`, `test_aggregate`, `test_summary` (ratio semantics), `test_encounter_rate` (mode following, overrides, units, presets, end-to-end through `summarize_df`).

### Notes for consuming workflows

- Purely additive — no existing public API changed.
- The flat-conditional form schemas are a deliberate departure from pydantic's default `anyOf` rendering; workflow specs may need rjsf-override updates to match the new labels ("Measure Encounters By", "Event Field to Sum").

Closes #802


ecoscope/platform/tasks/transformation/_indexing.py
```diff
@@ -144,6 +169,9 @@ def add_spatial_index(
     if not isinstance(groupers, AllGrouper):
         spatial_groupers = [g for g in groupers if isinstance(g, SpatialGrouper)]
         for sg in spatial_groupers:
+            if sg.index_name in gdf.index.names:
+                # Already indexed by an earlier call with an overlapping grouper list.
+                continue
```
atmorling, line 177: Interested to know why we skip in the case of overlapping indexes for spatial groupers, but for temporal groupers we overwrite an overlapping index? I'm not strongly opinionated one way or the other but we should have the behaviour be consistent across groupers. Perhaps adding a `skip_if_index_exists` flag or similar?

ecoscope/platform/tasks/analysis/_patrol_summary.py
```diff
@@ -186,3 +197,339 @@ def set_patrol_summary_metrics(
...
+    ] = "total_encounter"
+    aggregate_column: Annotated[str, _PER_METRIC_AGGREGATE_COLUMN_FIELD] = ""
+
+    def to_summary_param(self, aggregate_column: str | None = None) -> StatSummaryParam:
+        column = self.aggregate_column or aggregate_column
+        operand = _event_numerator(column)
+        # "Total Encounters" in count mode, not "Total Events": TotalEventsMetric
+        # owns that display name, and summarize_df keys columns by display name.
+        return StatSummaryParam(
+            display_name=f"Total {_event_label(column)}" if column else "Total Encounters",
```
atmorling, line 259: At the risk of a dumb question: why does `to_summary_param` allow overriding `aggregate_column`?

It seems like there's a convention to the display names based on the comment about "Total Events" being owned by a different model but there's nothing stopping me from calling `TotalEncounterMetric.to_summary_param("events")` right?

Mostly curious because, _if_ we removed that override then we could fix the display_name as a field on the model itself which would be a little more readable and less risky with regards to overlapping names.

atmorling (changes requested):
Looks great overall @Yun-Wu TYSM - have requested changes just to discuss the inconsistent indexing behaviour between `add_temporal_index` and `add_spatial_index`

ecoscope/platform/tasks/analysis/_patrol_summary.py
Yun-Wu: Good question. There may be multiple metrics in the patrol-encounter-rate-stats workflow and imho it's a bit tedious to put in the column name for every single one of them. So I created an extra field that will be applied to all encounter rate related metrics

ecoscope/platform/tasks/transformation/_indexing.py
Yun-Wu: Mainly because spatial groupers are almost guaranteed to generate the same result assuming there's one geometry column. But your comment makes sense. I've aligned them to both skipping duplicate indexes

atmorling (approved):

PR #805, opened 2026-08-03
Add `MCP` and `BBMM` home-range estimation methods

### LINKED ISSUE #804: Add `MCP` & `BBMM` home-range methods
ecoscope currently supports Elliptical Time-Density (ETD) for home-range estimation. This proposal adds two additional, well-established methods: Minimum Convex Polygon (MCP) and the Brownian Bridge Movement Model (BBMM).

### 1. Minimum Convex Polygon (MCP)

Minimum Convex Polygon (Mohr, 1947) is the oldest and simplest home-range estimator and remains widely used for exploratory analyses, small datasets, and comparisons with historical studies that reported home ranges exclusively as MCPs. Unlike ETD, MCP is a purely geometric method: it operates on an unordered set of location fixes and does not require a utilization distribution or spatial grid.

For each requested percentile:

* Compute the centroid of the complete set of locations.
* Rank locations by their distance from the centroid.
* Retain the closest fraction of points corresponding to the requested percentile, discarding the most distant locations as outliers.
* Compute the convex hull enclosing the retained points.
* Skip percentiles that retain fewer than three points, as a convex hull cannot be defined.

### 2. Brownian Bridge Movement Model (BBMM)

The Brownian Bridge Movement Model (Horne, Garton, Krone & Lewis, 2007) provides a statistically grounded alternative to ETD. Rather than treating each location independently, BBMM models the animal's probable movement between consecutive fixes as a Brownian bridge, producing a continuous utilization distribution similar to ETD while explicitly accounting for the movement path.

Key implementation details:

* Model each trajectory segment using the elapsed time, distance between fixes, GPS location error, and an estimated Brownian motion variance.
* Estimate the motion variance from the trajectory using a leave-one-out procedure over the interior fixes.
* Accumulate the probability density from all segments onto a shared spatial grid, then normalize the resulting utilization distribution so its total probability equals 1.
* Exclude segments separated by unusually large temporal gaps, as bridging long intervals produces unrealistically diffuse probability surfaces.

zakhusayn:
## :earth_americas: Summary

Closes #804 - adds Minimum Convex Polygon (MCP) and Brownian Bridge Movement Model (BBMM) as home-range estimation methods alongside the existing Elliptical Time-Density (ETD) method.

* **MCP:** The core algorithm (`calculate_mcp_range`) ranks fixes by their distance from the centroid, retains the closest fraction of points for each requested percentile, and returns the corresponding convex hull. It is a purely geometric method with no density surface. The task wrapper (`calculate_minimum_convex_polygon`) returns results in the same `percentile`/`geometry`/`area_sqkm` format used by ETD.
* **BBMM:** The core algorithm (`calculate_bbmm_range`) models the probability distribution of an animal's location between consecutive fixes as a function of elapsed time, distance, motion variance, and GPS location error, accumulating per-segment contributions onto a shared grid to produce a utilization-distribution surface. The motion variance is estimated from the trajectory itself using maximum likelihood. The task wrapper (`calculate_brownian_bridge_range`) returns results in the same output format as ETD and MCP.

* Added supporting tasks needed for MCP and BBMM to function as full peers of ETD in workflows:

  * `correct_non_overlapping_percentile_rings`, which prevents stacked semi-transparent choropleth layers from exhibiting compounded opacity where percentile polygons overlap.
  * `convert_trajectory_to_relocations`, which recovers point fixes from a trajectory, since MCP operates on point locations rather than trajectory segments.


ecoscope/platform/tasks/preprocessing/_preprocessing.py
```diff
@@ -153,3 +153,18 @@ def relocations_to_trajectory(
         raise ValueError("No Trajectory data left after applying segment filter")
 
     return traj.gdf
+
+
+@register()
+def convert_trajectory_to_relocations(trajectory_gdf: TrajectoryGDF) -> AnyGeoDataFrame:
```
Yun-Wu, line 159: why do we need this? Can we just use the output of `get_subjectgroup_observations` directly for MCP? Looks like you are also using `_extract_points`in BBMM? Is it possible to simplify these traj->reloc converstion?

ecoscope/platform/tasks/analysis/_raster.py
Yun-Wu, line 1: Is this module from -ext-custom? If so, let's also delete the file in -ext-custom

ecoscope/platform/tasks/analysis/_raster.py
```diff
@@ -0,0 +1,247 @@
...
+        dtype="float64",
+        nodata=etd_params["nodata_value"],
+    )
+
+    logger.info(f"ETD raster written to: {output_path}")
+    return output_path
+
+
+@register()
+def generate_bbmm_raster(
```
Yun-Wu, line 239: This task is very similar with `generate_etd_raster`. Can we try to merge them?

ecoscope/platform/tasks/config/_home_range.py
Yun-Wu, line 1: we can have a set of config tasks and use allOf/if/then to switch between branches to avoid showing unrelated config params, something like this: https://github.com/wildlife-dynamics/ecoscope/blob/master/ecoscope/platform/tasks/analysis/_patrol_summary.py#L450

ecoscope/analysis/UD/bbmm_range.py
```diff
@@ -0,0 +1,231 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _extract_points(trajectory_gdf: gpd.GeoDataFrame) -> tuple[np.ndarray, np.ndarray]:
+    """Reconstruct the point sequence (xy, t) from a trajectory's LineString
+    segments, reusing `Trajectory.to_relocations()` - the same canonical
+    trajectory->points conversion `convert_trajectory_to_relocations` uses for
+    MCP - rather than a separate reimplementation. `t` is UNIX seconds (float)."""
+    relocs_gdf = Trajectory(gdf=trajectory_gdf).to_relocations().gdf
```
Yun-Wu, line 25: does it work for multiple subjects?

ecoscope/analysis/UD/bbmm_range.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def estimate_motion_variance(trajectory_gdf: gpd.GeoDataFrame, location_error: float) -> float:
+    """MLE for a single, whole-track sigma_m^2 (Horne et al. 2007, Eqn 1)."""
+    midpoints = _compute_midpoints(trajectory_gdf)
+    if midpoints.empty:
+        raise ValueError("Not enough interior fixes with valid time gaps to estimate motion variance.")
+    result = minimize_scalar(
+        _neg_log_likelihood,
+        bounds=(1.0, 1_000_000.0),
```
Yun-Wu: have you checked this bound? Based on Claude 1.0 is ~85 m per hour and 1_000_000 is about 85 km per hour.  A tortoise, a denning bear, or a bird on the nest can go below 1 m² s⁻¹ and a soaring bird in migration exceeds 10⁶. At least we should warn on these scenarios

ecoscope/analysis/UD/bbmm_range.py
zakhusayn: It's a helper func, so it just processes whatever `trajectory_gdf` it's given (it doesn't group by subject itself). It works for multiple subjects only because the workflow always calls it once per subject already-split trajectory.

ecoscope/analysis/UD/bbmm_range.py
zakhusayn: Good catch, you're right that the bound can clip real cases. The optimizer we use doesn't complain when the true answer falls outside its search range, it just quietly hands back whichever edge is closer, with nothing telling you that happened.

I added a warning instead of just widening the numbers, since picking new bounds would've been just as random without real data on how still or fast-migrating animals actually look here. Now if the estimate lands right on one of the edges, it tells you which one and what that probably means, barely moving on the low end, moving very fast on the high end, so you know to double-check the result instead of trusting it blindly. Added tests for both cases too.

ecoscope/platform/tasks/analysis/_raster.py
zakhusayn: Good point, there was real duplication there. I pulled out the shared part, the empty-data check, writing the file, logging, and returning the path, into one helper function that both tasks now call.

I kept them as two separate tasks though. They take different inputs, they're wired to different steps in the workflow that only run depending on which method the user picked, and the actual math each one does is genuinely different, not just organized differently. ETD has to work out its own grid size before it can run, while BBMM just passes its settings straight through. Merging them into one task would mean reworking how the workflow decides which one to run, which is a bigger change than the duplication really calls for.

ecoscope/platform/tasks/config/_home_range.py
zakhusayn: Thanks for the suggestion, we actually tried this one already. It broke on defaults, since RJSF fills in the form's starting values before it works out which if/then branch applies, so any default set inside a branch just showed up blank.

The current setup already hides unrelated fields per method, and gives us defaults for free since each method is its own plain model. Since it already works and this would trade that away, we're keeping it as is.

ecoscope/platform/tasks/preprocessing/_preprocessing.py
zakhusayn: Because MCP needs to run on the exact same filtered, grouped, per-subject point set that ETD and BBMM.

1. `subject_traj_groups` builds the Trajectory once per subject, after all the upstream Filter data/group data steps have already run, and that same per-subject split feeds ETD, BBMM, and MCP identically via mapvalues. If MCP pulled from `get_subjectgroup_observations` directly instead, it would skip that shared filtering/grouping pipeline and could end up computing its home range over a different set of points than ETD/BBMM saw breaking the guarantee that all three methods answer the same question, just via different algorithms. 

2. On BBMM's `_extract_point` I've simplify that: it now calls the same `Trajectory.to_relocations()` method `convert_trajectory_to_relocations` uses internally, instead of a separate reimplementation of the same point-reconstruction logic. So both MCP and BBMM go through the identical canonical trajectory -> points conversion now.

ecoscope/analysis/UD/bbmm_range.py
Yun-Wu: Will the warning be exposed to our users? If not, how about adding them as a configuration option?

Yun-Wu (approved):
Looks great! Thank you very much Zak!

PR #807, opened 2026-08-03
Allow `gather_dashboard` to accept a list of file paths

### LINKED ISSUE #774: Allow `gather_dashboard` to accept a list of file paths to include in its output
Beyond adding the input arg, this will mean:
- Updating the dashboard [base model](https://github.com/wildlife-dynamics/ecoscope/blob/29bcc087586eeecbbdec545268aff935992c57a9/ecoscope/platform/tasks/results/_dashboard.py#L95-L106) to serialise the input files by group 
- Adding analogue tasks to `create_*_widget_single_view` and `merge_widget_views` in order to handle grouped file lists at the spec level

atmorling:
## :earth_americas: Summary
Closes #774 

## :package: Proposed Changes
- Adds a `files` arg to `gather_dashboard` which allows workflows to declare files as formal outputs of the workflow (as opposed to being something that implicitly happens or doesn't happen as the result of an upstream task)
  - If `files` is `None` the dashboard will serialise to the existing contract
  - If `files` is a list, empty or otherwise, the dashboard will serialise to the new contract:
```diff
        "views": {
+           "{\"All\": \"True\"}": {
+               "dashboard": [
                    {
                        "widget_type": "graph",
                        "title": "Events Bar Chart",
                        "is_filtered": true,
                        "id": 0,
                        "data": "/tmp/results/90e9447_v2.html"
                    }
+               ],
+               "files": [
+                   {
+                       "path": "/tmp/results/my_events.parquet"
+                   }
                ]
+           }
        },
```

- Adds equivalent widget tasks and types to service creating lists of file paths to feed to the dashboard
- I've tested these changes against:
  - The [events workflow](https://github.com/ecoscope-platform-workflows-releases/events/pull/212) for regression
  - A test workflow demonstrate the new contract [here](https://github.com/atmorling/test-dashboard-workflow/blob/9e635e9ce91eff40c2a6b4b22ad46fac7b1698f9/__results_snapshots__/default/test_dashboard_json%5Bdata%5D.json)

## 📋 Checklist

### DIFF
--- a/ecoscope/platform/tasks/results/_dashboard.py
+++ b/ecoscope/platform/tasks/results/_dashboard.py
@@ -93,17 +93,27 @@ class DashboardFile(BaseModel):
 
 
 class DashboardView(BaseModel):
-    """The contents of a single dashboard view: its widgets and associated files."""
+    """The contents of a single dashboard view: its widgets and associated files.
+
+    This wrapper is only used when a dashboard actually has associated files. When a
+    dashboard has no files, each view is serialized as a bare list of widgets (see
+    `Dashboard._iter_views_json`) to preserve the historical output shape.
+    """
 
     dashboard: list[EmumeratedWidgetSingleView]
     files: list[DashboardFile]
+
+
+# A view is serialized either as the file-aware `DashboardView` (when the dashboard has
+# files) or as a bare list of widgets (the historical shape, when it has none).
+SerializedView = DashboardView | list[EmumeratedWidgetSingleView]
 
 
 class DashboardJson(BaseModel):
     """A JSON-serialized representation of a dashboard."""
 
     filters: dict | None
-    views: dict[str, DashboardView]
+    views: dict[str, SerializedView]
     metadata: Metadata
     layout: list  # this is a placeholder for future use by server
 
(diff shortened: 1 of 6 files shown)


ecoscope/platform/tasks/results/_dashboard.py
```diff
@@ -316,6 +379,10 @@ def gather_dashboard(
     ] = None,
     time_range: Annotated[TimeRange | SkipJsonSchema[None], Field(description="Time range filter")] = None,
     warning: Annotated[str | SkipJsonSchema[None], Field(exclude=True)] = None,
+    files: Annotated[
+        NestedFileList | FlatFileList | GroupedOrSingleFile | SkipJsonSchema[None],
+        Field(description="Files to associate with the dashboard views.", exclude=True),
+    ] = None,
```
atmorling, line 385: This defaults to the legacy contract (no files field), which is semantically the same as an empty files list in the new contract

ecoscope/platform/tasks/results/_widget_tasks.py
```diff
@@ -23,6 +26,15 @@ def _fallback_to_none(obj: WidgetData | SkipSentinel) -> WidgetData | None:
     return None if isinstance(obj, SkipSentinel) else obj
 
 
+def _fallback_to_empty_list(obj: Path | list[Path] | list[SkipSentinel] | SkipSentinel) -> Path | list[Path]:
+    """Fallback function to convert SkipSentinel to an empty list."""
+    if isinstance(obj, SkipSentinel) or (
+        isinstance(obj, list) and all([isinstance(item, SkipSentinel) for item in obj])
+    ):
+        return []
+    return cast(Path | list[Path], obj)
```
atmorling, line 35: mypy doesn't see that the above check obviates and return of `SkipSentinel`

ecoscope/platform/tasks/results/_dashboard.py
```diff
@@ -83,11 +86,34 @@ def ser_model(self) -> dict[str, Any]:
...
+# files) or as a bare list of widgets (the historical shape, when it has none).
+SerializedView = DashboardView | list[EmumeratedWidgetSingleView]
+
+
 class DashboardJson(BaseModel):
     """A JSON-serialized representation of a dashboard."""
 
     filters: dict | None
-    views: dict[str, list[EmumeratedWidgetSingleView]]
+    views: dict[str, SerializedView]
```
cisaacstern, line 116: nice 👏

ecoscope/platform/tasks/results/_dashboard.py
```diff
@@ -140,25 +169,51 @@ def _get_view(self, view: CompositeFilter | None) -> list[EmumeratedWidgetSingle
             for i, w in enumerate(self.widgets)
         ]
 
+    def _get_files(self, view: CompositeFilter | None) -> list[DashboardFile]:
+        """Get the files associated with a view. Files keyed by `None` are ungrouped/global:
+        they apply to every view, so they are included in addition to any files keyed to the
+        requested `view`. This means a dashboard can combine ungrouped files (a single `None`
+        key) with grouped files (per-view keys) and both surface in each view.
```
cisaacstern, line 176: From a design standpoint, would we want to at a property to these file entries so that the UI can somehow indicate that these are "global" and not view-specific?

ecoscope/platform/tasks/results/_dashboard.py
```diff
@@ -83,11 +86,34 @@ def ser_model(self) -> dict[str, Any]:
         return all_fields
 
 
+class DashboardFile(BaseModel):
+    """A file associated with a dashboard view."""
+
+    path: Path
```
cisaacstern, line 92: would we potentially want any other metadata tied the the DashboardFile aside from path? title / description / size / etc. ?

cisaacstern (approved):
lgtm, some musings about future directions inline

ecoscope/platform/tasks/results/_dashboard.py
atmorling: The behaviour here is consistent with how the widgets work today in that, ungrouped (view=None) widgets are surfaced in every view. Adding a flag _in general_ seems good in both cases because it removes the need for repeating properties, but represents a bigger, contract change from the frontend's pov than the one scoped here which is purely 'additive'.

ecoscope/platform/tasks/results/_dashboard.py
atmorling: Yup in future for sure. IMO this is out of scope for Right Now :tm: as we don't have a clear story around what metadata should be included and how this would surface in the frontend

PR #814, opened 2026-08-10
Break multi-day trajectory segments into individual solar day segments in `get_nightday_ratio`

### LINKED ISSUE #812: Implement fix for night day ratio inaccuracy
There's two components to this:
- One is the current ratio is actually an average of each individual calendar day's ratio, which allows for a single bad day's worth of data (where the ratio is an outlier) to disproportionally influence the final ratio. We probably want to be calculating the ratio based on distance travelled night / distance travelled day for the _entire_ dataset given.
- Our internal daily summary calculation (and therefore the day/night fractions that are derived from it), do not properly account for trajectory segments than span multiple days. We should pre-process the input trajectories to carve them up into individual calendar days.

atmorling:
## :earth_americas: Summary
Closes #812 

## 📋 Checklist


ecoscope/analysis/astronomy.py
```diff
@@ -129,90 +129,162 @@ def calculate_day_night_distance(
...
 
-    daily_summary["night_day_ratio"] = daily_summary["night_distance"] / daily_summary["day_distance"]
-    mean_night_day_ratio = daily_summary["night_day_ratio"].replace([np.inf, -np.inf], np.nan).dropna().mean()
-    return mean_night_day_ratio
+    # Map the mean fraction back onto a ratio scale, so 1.0 still means balanced
+    # movement and the returned value stays comparable to the historical ratio. A fully
+    # nocturnal track (mean fraction 1.0) maps to an infinite ratio.
+    mean_fraction = night_fraction.mean()
+    if mean_fraction >= 1.0:
+        return np.inf
```
atmorling: Worth noting that `np.inf` would coerce to `"inf"` in a single value widget

tests/conftest.py
```diff
@@ -1,17 +1,67 @@
...
+    ``bearing_deg`` (0 = north, 90 = east), keeping the geometry self-consistent with the
+    stated distance rather than using an arbitrary fixed step. Assemble a list of these into
+    a GeoDataFrame with ``build_segments``.
+    """
+
+    start_point: Point
+    start_time: pd.Timestamp
+    end_time: pd.Timestamp
+    dist_meters: float = 1000.0
+    bearing_deg: float = 45.0
```
atmorling, line 36: My hope with this is to enable more robust testing of specific types of movement data by giving us a way to declare a trajectory segment with specific attributes

ecoscope/analysis/astronomy.py
```diff
@@ -129,90 +129,162 @@ def calculate_day_night_distance(
...
 def get_nightday_ratio(gdf: gpd.GeoDataFrame) -> float:
+    """Night/day movement ratio, averaged per local solar day.
+
+    Calculate the night fraction for each local solar day, ``night / (night + day)``, and the
+    mean of those fractions across days is mapped back to a ratio via (``mean_fraction / (1 - mean_fraction)``)
+    1.0 means balanced movement, >1 more nocturnal, <1 more diurnal.
+
+    Multi-day segments are first split at local-midnight boundaries (see
+    ``_segments_by_local_day``) so a single long gap-bridging segment is apportioned across
+    the days it spans instead of being scored entirely against one day's sunrise/sunset.
```
Yun-Wu: Is the night day ratio degraded to the night/day local timespan ratio for a multi-day segment, unrelated to the movement itself? Is it a valid hypothesis? Say a bear only moves at night and for a really long distance in the summer, when it happens to have a long daytime. Then the night/day ratio could be really low and pollute the actual result.

Yun-Wu (changes requested):
Thanks @atmorling for addressing this. Please see my comments below. I wonder if mult-day segment should be included in calculating night/day ratio as it does not contribute to any meaningful information.

Also, there are a few datasets in the er-asia instance: elephants/snow leopard/vultures. Could you test the algorithm against them to make sure it doesn't break the existing flows?

ecoscope/analysis/astronomy.py
atmorling: It's a good question - a couple things worth mentioning:
- Multi-day segments can occur _very_ easily because we build tracks fix-to-fix, so if the collar/tag data are spotty it's easy to end up with a big temporal jump in the track.
- The current algorithm (and the one in this pr), calculate the mean of each individual day's ratio, which is sensitive to pollution from a single day outlier. I've synced with @walljcg on this and this is an accepted trade-off.

Ultimately the rationale for breaking up multi-day segments here is to reduce the potential for pollution in the ratio, however there will always be cases like your bear example, or even simply tracks built from bad fix data, where the upstream trajectory segment filters would need to be tuned. This change definitely assumes that you have filtered your data down to just the tracks that "make sense".

I'm open to changing this to explicitly exclude any multi-day segments entirely (instead of splitting them), but my slight preference is for this calculation to be not heavily opinionated.

walljcg (approved):
Looks good to me! Thanks for the extensive re-work.

PR #823, opened 2026-08-12
ERDW-276: Fix DWH get_events failure when event details are requested

### LINKED ISSUE #822: Download Events fails with DWH
### Steps to reproduce

Create a Download Events workflow and run it using the data warehouse from advanced run settings

### Expected Results

The workflow should run to completion without issues.

### Actual Results

The workflow fails with error:
```bash 
Task instance 'get_event_data' raised ValueError('include_details (and parse_detail_datetimes / invalid_only / invalid_details) require exactly one event_type, since the typed event_details struct is derived from a single event type's schema. Pass raw_details=True for raw JSON event_details across multiple event types.')`
```

### Acceptance Criteria

_No response_

### Notes

_No response_

marianobrc:
## :earth_americas: Summary
Closes #822 

## :package: Proposed Changes
The DWH returns typed `event_details` as an Arrow struct by default, which requires exactly one event type, so `get_events` raised whenever  `include_details` was set with zero or many types, breaking the  `download-events` workflow.

The task now picks the mode by selection count: the typed struct for a single  event type (`from_arrow` lands it as dicts), and the raw JSON form, decoded to  dicts, for zero or many. Either way the column matches what the ER API path serves, so no workflow changes are needed.

### DIFF
--- a/ecoscope/platform/tasks/io/_earthranger.py
+++ b/ecoscope/platform/tasks/io/_earthranger.py
@@ -722,6 +722,12 @@ def get_events(
 
         # The warehouse get_events accepts event type slugs directly, so pass the
         # values straight through — no value->id round-trip via get_event_types.
+        #
+        # raw_details: the warehouse can also serve `event_details` as structured Arrow
+        # data, but only for a single event type (the struct is derived from that type's
+        # schema), and adopting it would mean refactoring the tasks and workflows that
+        # consume the column, which expect dicts. So ask for the raw JSON form — valid
+        # for any number of event types — and decode it to dicts below.
         table = warehouse_client.get_events(
             since=time_range.since.isoformat(),
             until=time_range.until.isoformat(),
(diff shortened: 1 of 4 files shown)


atmorling (approved):

ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -722,6 +722,14 @@ def get_events(
...
         # The warehouse get_events accepts event type slugs directly, so pass the
         # values straight through — no value->id round-trip via get_event_types.
+        #
+        # event_details: the warehouse serves it as a structured Arrow struct, which
+        # `from_arrow` lands as dicts, but that struct is derived from one event type's
+        # schema and so requires a single-type selection. Workflows also allow zero or
+        # many types (download-events offers "leave empty for all event types"), so fall
+        # back to the raw JSON form there and decode it below — either way the column
+        # ends up dicts, as the ER API path serves it.
+        raw_details = include_details and len(event_types) != 1
```
marianobrc, line 732: @atmorling I requested a new review since I pushed a change in this line (`and len(event_types) != 1`) so that typed mode is used if a single event type is selected, falling back to raw otherwise

atmorling: Thanks! Makes sense

atmorling (approved):
Thanks @marianobrc

PR #829, opened 2026-08-27
Skip overwrite on content hashed persistence tasks

### LINKED ISSUE #826: Skip writing the file in the event of a naming collision in persist tasks where a content hash is used
A naming collision on a content-hashed file means we'd be overwriting with identical data, so let's skip the write in these cases

atmorling:
## :earth_americas: Summary
Closes #826 

## :package: Proposed Changes
- Adds a check to the serde module to check if a file already exists or not
- Wires the above check into file persistence tasks where a content hash is used to generate the filename, if there's a collision we will no longer write anything
- Factoring changes to `_persist_df` for better readability
- Some improvements to`_hash_df` to coerce unhashable types

## 📋 Checklist

### DIFF
--- a/ecoscope/platform/tasks/io/_earthranger.py
+++ b/ecoscope/platform/tasks/io/_earthranger.py
@@ -24,7 +24,7 @@ from ecoscope.platform.schemas import (
     SpatialFeaturesGroup,
     SubjectGroupObservationsGDF,
 )
-from ecoscope.platform.serde import _persist_bytes
+from ecoscope.platform.serde import _persist_bytes, _read_path_if_file_exists
 from ecoscope.platform.tasks.filter._filter import TimeRange
 from ecoscope.platform.tasks.io._persist import (
     _fallback_to_empty_grouped,
@@ -1253,6 +1253,9 @@ def download_grouped_event_attachments(
     `split_groups` — and writes each event's files to
     `output_dir/<attachments_subdir>/<key_hash>_<event_id>/<filename>`.
 
+    A file already present at the target path is left alone: it is neither
+    refetched nor rewritten.
+
     `<key_hash>` is a short sha256 hash of the encoded group key, computed
     by the same helper used by `persist_grouped_dfs_for_results_download`,
     so the FE can match downloaded attachments to the dashboard view they
@@ -1310,12 +1313,18 @@ def download_grouped_event_attachments(
                 filename = result["filename"]
                 basename = Path(filename).name
 
-                response = client._get(  # type: ignore[attr-defined]
-                    f"activity/event/{event_id}/file/{file_id}/original/{filename}",
-                    return_response=True,
-                )
+                # Each event has its own folder, so a name clash here means a
+                # second attachment on this event sharing a filename. Keep the
+                # first: overwriting it would fail anyway on a `gs://` root,
+                # where replacing an object needs a delete permission the
+                # workflow service account does not have.
+                if _read_path_if_file_exists(event_dir, basename) is None:
+                    response = client._get(  # type: ignore[attr-defined]
+                        f"activity/event/{event_id}/file/{file_id}/original/{filename}",
+                        return_response=True,
+                    )
 
-                _persist_bytes(response.content, event_dir, basename)
+                    _persist_bytes(response.content, event_dir, basename)
 
                 downloaded_paths.append(f"{attachments_subdir}/{event_dir_name}/{basename}")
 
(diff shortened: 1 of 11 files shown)


ecoscope/platform/tasks/io/_earthranger.py
```diff
@@ -1310,12 +1313,17 @@ def download_grouped_event_attachments(
...
-                )
+                # Each event has its own folder, so a name clash here means a
+                # second attachment on this event sharing a filename.
+                # Shouldn't be possible due to upstream conventions, but
+                # handling defensively here by skipping the duplicated write.
+                if _read_path_if_file_exists(event_dir, basename) is None:
+                    response = client._get(  # type: ignore[attr-defined]
+                        f"activity/event/{event_id}/file/{file_id}/original/{filename}",
+                        return_response=True,
+                    )
```
atmorling, line 1324: IIUC the upstream API prevents this, but better to be defensive as there's nothing in the contract we consume that guarantees the uniqueness of an attachment filename.

ecoscope/platform/tasks/io/_persist.py
```diff
@@ -109,26 +123,34 @@ def persist_json(
...
 
-    # Use a hash of the dataframe content; avoids issues with unhashable types.
+    def _fallback_hash(obj):
+        try:
+            return pd.util.hash_pandas_object(obj, index=False).values
+        except (TypeError, ValueError, AttributeError):
+            # Cells pandas can't hash (lists/dicts in event details). `repr`
+            # rather than `str`: in a mixed column `str` collapses 1 and "1"
+            # to the same text, and a filename collision serves wrong bytes.
+            return pd.util.hash_pandas_object(obj.map(repr), index=False).values
```
atmorling: Previously we hashed just the shape + first 5 rows of the df in the event of any unhashable types, now we're coercing anything unhashable so we can consider the whole frame.

ecoscope/platform/tasks/io/_persist.py
atmorling, line 1: The big diff here is mostly due to refactoring

ecoscope/platform/tasks/io/_persist.py
```diff
@@ -146,41 +168,106 @@ def persist_df(
...
-        gdf = gpd.GeoDataFrame(df)
-        return _persist_text(gdf.to_json(), root_path, f"{filename}.{filetype}")
-    elif filetype == "json":
-        return _persist_text(df.to_json(), root_path, f"{filename}.{filetype}")
-    else:
-        raise ValueError(f"Unsupported file type: {filetype}")
+    target = f"{filename}.{_filetype_extension(filetype)}"
+
+    if filetype == "geoparquet":
+        _require_geometry(df, filetype)
```
atmorling, line 249: Other than skipping duplicated writes, this is the only other logic change in this module, if one attempt to persist a geoparquet without geometry we'll now explicitly `ValueError`

ecoscope/platform/tasks/io/_persist.py
```diff
@@ -295,16 +372,16 @@ def persist_df_wrapper(
     else:
         df_new = df
 
-    filehash = _hash_df(df)
+    filehash = _hash_df(df_new)
```
atmorling, line 393: Previously this hash was built from the pre-sanitized df, we're now hashing what we intend to write.

ecoscope/platform/tasks/io/_persist.py
```diff
@@ -295,16 +372,16 @@ def persist_df_wrapper(
...
 
-    paths = [persist_df(df_new, root_path, filename, filetype) for filetype in filetypes]
+    paths = [_persist_df(df_new, root_path, filename, filetype, content_addressed=True) for filetype in filetypes]
 
     return paths
 
 
 def _hash_grouper_key(composite_filter: CompositeFilter) -> str:
-    """Return a 7-char sha256 hash of the JSON-encoded grouper key.
+    """Return a 6-char sha256 hash of the JSON-encoded grouper key.
```
atmorling, line 402: Changes here to correct a subtlety with the grouper key hash - thankfully this will be deprecated soon

ecoscope/platform/tasks/io/_persist.py
Yun-Wu: This makes sense as most of our unhashable objects are list/dict. Should we add a comment here that it won't be guaranteed for custom data classes if repr is overridden in the class and is not guaranteed to be unique?

ecoscope/platform/tasks/io/_persist.py
Yun-Wu: Thanks for catching this!

Yun-Wu (approved):

ecoscope/platform/tasks/io/_persist.py
atmorling: I've updated so that we explicitly raise in the case where the repr fails

PR #830, opened 2026-08-27
Remove unsafe .join in `apply_sql_query` and `persist_df_wrapper`

### LINKED ISSUE #827: Remove unsafe `.join` in `apply_sql_query` and `persist_df_wrapper`
The use of pandas.join in these tasks is unsafe against dataframes that have duplicate indexes, which is possible with _at least_ patrols datasets

atmorling:
## :earth_americas: Summary
Closes #827 

## :package: Proposed Changes
- Updates `persist_df_wrapper` and `apply_sql_query` to reattach geometry positionally instead of via `.join`, in the case of duplicate indexes `.join` duplicates data in the frame

## 📋 Checklist


Yun-Wu (approved):
Thanks for detecting this issue this Alex!

PR #836, opened 2026-09-04
Allow users to define how `map_columns` handles duplicates

### LINKED ISSUE #835: Allow users to define how `map_columns` handles duplicates
Allow callers to set whether to skip, overwrite or raise when `map_columns` encounters duplicates. Currently the rename logic does not gracefully handle duplicates at all

atmorling:
## :earth_americas: Summary
Closes #835 

## :package: Proposed Changes
- Adds a `_safe_rename_columns` utility shared by tasks in `_mapping` that rely on `df.rename`
- Adds a `duplicate_strategy` arg to `map_columns` allowing users to chose either "suffix", "skip", "overwrite", or "error".
  - Uses "prefix" as the default as that seems safest in the average case, though much of our consuming workflows will want "overwrite".
- Additionally to `map_columns`, `duplicate_strategy` is also wired up as args on `title_case_columns_by_prefix` and `strip_prefix_from_column_names`

## 📋 Checklist

### DIFF
--- a/ecoscope/platform/tasks/transformation/_mapping.py
+++ b/ecoscope/platform/tasks/transformation/_mapping.py
@@ -1,6 +1,8 @@
 import logging
-from typing import Annotated, Literal, cast
+from collections.abc import Collection
+from typing import Annotated, Literal, cast, get_args
 
+import pandas as pd
 from pydantic import BaseModel, Field
 from pydantic.json_schema import SkipJsonSchema
 from wt_registry import register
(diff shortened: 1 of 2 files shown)


atmorling:
@Yun-Wu for visibility I realised while working on this that it re-treads a lot of ground that you've covered in `_column_cleanup` - have opened https://github.com/wildlife-dynamics/ecoscope/issues/835 to track aligning these two things

ecoscope/platform/tasks/transformation/_mapping.py
```diff
@@ -104,6 +105,167 @@ def format_row(x):
...
+
+    if to_drop:
+        # Dropping the active geometry column leaves a GeoDataFrame with no geometry, which
+        # demotes it to a plain DataFrame and loses the CRS.
+        if (active_geometry := _active_geometry_name(df)) in to_drop:
+            raise ValueError(
+                f"Renaming columns {mapping} would overwrite the active geometry column "
+                f"'{active_geometry}'. Drop or rename it explicitly first, or use a "
+                "duplicate_strategy of 'skip', 'suffix' or 'error'."
+            )
```
atmorling, line 260: Explicitly prevent dropping the active geometry column. It should never be the expected outcome from tasks in this module.

ecoscope/platform/tasks/transformation/_mapping.py
```diff
@@ -104,6 +105,167 @@ def format_row(x):
...
+    owner: dict[str, str] = {col: col for col in df.columns if col not in mapping}
+    resolved: dict[str, str] = {}
+    to_drop: list[str] = []
+    for old, new in mapping.items():
+        if new in owner:
+            displaced = owner.pop(new)
+            to_drop.append(displaced)
+            resolved.pop(displaced, None)
+        resolved[old] = new
+        owner[new] = old
```
atmorling, line 250: Overwrite is useful when prepping for display, in these cases we almost never surface anything other than the columns we're trying to map, so dropping the original column is ideal

